PR #19, opened 2016-04-18
Check for hook existence at install/uninstall

vanengers:


RicardoJ:
Thanks George. I also recieved your email. I'll fix the tests before I will merge this within master and I'll then release a new version.

PR #29, opened 2016-12-07
Add PrestaShop 1.7 support

firstred:
Add support for PrestaShop 1.7

In order to achieve this, a few changes were necessary.

Properties such as `name`, `author`, `tab`, etc. had to be moved to the constructor (due to module file parsers). This also allows you to check if the module is PS1.7 compatbile via https://beta-validator.prestashop.com

The dependency on `blockcart` has been removed. The availability of this module cannot be guaranteed and shouldn't be necessary for a payment module.

Instead of implementing the new `paymentOptions` hook, this PR recycles the `displayPaymentEU` hook, since we do not need the new fancy forms. This is what it looks like with `displayPaymentEU`:
mollie17

`$cookie` should be accessed via the `Context` object (`$this->context` or `Context::getContext`).

`css` folder had to be moved to `views/css/` to fix the file structure of the module.

`Module::getPaymentModules` on 1.7 does not return modules which only support the `displayPaymentEU` hook (bug?). A targeted check for `displayPaymentEU` has been added in order to correctly check if the current user is authorized.

Alrighty, this should be all you need to get the module working on 1.7. Enjoy!

[v1.3.0-mollie.zip](https://github.com/mollie/Prestashop/files/637263/v1.3.0-mollie.zip)


staeren:
get "unable to install mollie" on prestashop 1.7.0.2

xBorderie:
Hi Mollie team! @firstred is a knowledgeable member of the PrestaShop community, and therefore I am confident that this PR does what is needed for the module to work with PS 1.7.

nvdk:
I tried this branch (by running release.sh and uploading the resultant zip to /modules) and got the following error:

```
PHP message: PHP Fatal error:  require_once(): Failed opening required '/modules/mollie/lib/src/Mollie/API/Autoloader.php' (include_path='/tools/htmlpurifier/standalone:/vendor/pear/pear_exception:/vendor/pear/console_getopt:/vendor/pear/pear-core-minimal/src:/vendor/pear/archive_tar:.:/usr/share/php:/usr/share/pear') in /modules/mollie/mollie.php on line 103" while reading response header from upstream, client: 62.235.32.145, server: example.com, request: "GET /admin/index.php/module/manage?_token=cf-mitty2mT4iaTYsfjHPx3cePSk3iF98sjzA0JAPdA HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5-fpm.sock:", host: "example.com", referrer: "http://example.com/admin/index.php/module/catalog?_token=cf-mitty2mT4iaTYsfjHPx3cePSk3iF98sjzA0JAPdA"
```

Using PrestaShop version: 1.7.0.3

nvdk:
Tested on prestashop 1.7.0.3, got the following failure in prestashop log on install: Mollie::__construct said: Payment error:Invalid API key: \'\'. An API key must start with \'test_\' or \'live_\'.	

(also note for inexperienced testers like myself, you need a recursive clone of the repository as `/lib` is a git submodule)

nvdk:
After some mistakes on my part, I managed to install the ps17support branch and can confirm it seems to be working on prestashop 1.7.0.3.

joskepeeters:
Hi, i installed this 1.7module in my Prestashop and i can configure all the details but the payment option is not visible as a payment method and not showing up on the parments screen when ordering.
Is there a solution for this?

PR #43, opened 2018-02-27
Fix for missing template error

bastiaantenklooster:
Prestashop 1.7.2.4 will report that the mollie_return.tpl file is missing when not using module: for the filename. This causes an error in development mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ views/js/app/dist/**
 views/js/app/node_modules/**
 yarn.lock
 package-lock.json
+.idea/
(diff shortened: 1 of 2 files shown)


bastiaantenklooster:
I was not able to test if the same applies to the error.tpl file in the front/payment.php controller. It may be worth testing for that as well.

snowycatsoftware:
Hi Bastiaan,

Thank you for taking the time to notify us about this. The same must apply for error.tpl. We will run some tests and fix this issue in the next update.

Kind regards

snowycatsoftware:
All good. Thank you!

Issue #50, opened 2018-04-11
'Send webshop locale' setting results in redirect to homepage 


preitsma:
Mollie Prestashop Addon 2.0.4
Prestashop 1.6.1.10 
Module setting: send locale for payment screen set to 'Send webshop locale'

Clicking Ideal in the final screen of the Checkout Flow redirects me the home page instead of the Mollie Bank selection screen.
This happens on Safari and Firefox, not on Chrome. Did not test on IE.

snowycatsoftware:
Hi @preitsma ,

We just release the new version with the fix for this issue.

Kind regards

PR #51, opened 2018-04-11
Fix webshop locale passed to payment screen

firstred:
This PR fixes the webshop locale that is passed to the payment screen. Used when setting the `
Send locale for payment screen` to `Send locale for payment screen`


firstred:
Fixes #50

PR #137, opened 2019-10-22
Support

margud:


TomanDesign:
Hi, there is a problem with iDEAL payments. After the customer makes the payment, no tax is added to the invoice summary. There is an overpayment in the payment area.

PR #499, opened 2022-01-06
MOL-534: Fixed issue where order product price would be incorrect if …

margud:
…it didn't use default currency


lukvai (approved):

GytisZum (approved):
Looks good to me

PR #659, opened 2023-01-31
added customer page controller where you can see your subscriptions a…

margud:
Added subscriptions list in front page with subscription page, where you can see information about your subscriptions and change its method.
image
image
image

### DIFF
--- a/subscription/Handler/SubscriptionPaymentMethodUpdateHandler.php
+++ b/subscription/Handler/SubscriptionPaymentMethodUpdateHandler.php
@@ -37,9 +37,11 @@ class SubscriptionPaymentMethodUpdateHandler
     {
         $molPayment = $this->paymentApi->getPayment($transactionId);
         $recurringOrder = $this->recurringOrderRepository->findOneBy(['mollie_subscription_id' => $subscriptionId]);
+
         if (!$recurringOrder) {
             throw new MollieException('todo');
         }
+
         $subscriptionUpdateData = $this->subscriptionDataFactory->build($recurringOrder, $molPayment->mandateId);
         $newSubscription = $this->subscriptionApi->updateSubscription($subscriptionUpdateData);
         $recurringOrder->payment_method = $molPayment->method;
--- a/subscription/Handler/SubscriptionPaymentMethodUpdateHandler.php
+++ b/subscription/Handler/SubscriptionPaymentMethodUpdateHandler.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mollie\Subscription\Handler;
+
+use Mollie\Exception\MollieException;
+use Mollie\Subscription\Api\PaymentApi;
+use Mollie\Subscription\Api\SubscriptionApi;
+use Mollie\Subscription\Factory\UpdateSubscriptionDataFactory;
+use Mollie\Subscription\Repository\RecurringOrderRepositoryInterface;
+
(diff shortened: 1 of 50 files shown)


subscription/Handler/SubscriptionPaymentMethodUpdateHandler.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->recurringOrderRepository = $recurringOrderRepository;
+        $this->subscriptionDataFactory = $subscriptionDataFactory;
+        $this->paymentApi = $paymentApi;
+    }
+
+    public function handle(string $transactionId, string $subscriptionId)
+    {
+        $molPayment = $this->paymentApi->getPayment($transactionId);
+        $recurringOrder = $this->recurringOrderRepository->findOneBy(['mollie_subscription_id' => $subscriptionId]);
+        if (!$recurringOrder) {
```
mandan2, line 40: line spacing before and after if

subscription/Presenter/RecurringOrderPresenter.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function present(int $recurringOrderId): array
+    {
+        $recurringOrder = $this->recurringOrderRepository->findOneBy(['id_mol_recurring_order' => $recurringOrderId]);
+        $recurringProduct = $this->recurringOrdersProductRepository->findOneBy(['id_mol_recurring_orders_product' => $recurringOrderId]);
+
+        $product = new Product($recurringProduct->id_product, null, $this->language->getDefaultLanguageId());
+        $combination = new Combination($recurringProduct->id_product_attribute, null, $this->language->getDefaultLanguageId());
+        $order = new Order($recurringOrder->id_order);
+        $currency = new Currency($order->id_currency);
```
mandan2, line 48: spacing before new array instance

subscription/Repository/RecurringOrdersProductRepository.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace Mollie\Subscription\Repository;
+
+use MolRecurringOrdersProduct;
+use PrestaShopCollection;
+
+class RecurringOrdersProductRepository extends AbstractRepository implements RecurringOrdersProductRepositoryInterface
+{
+    public function findOneBy(array $keyValueCriteria): ?MolRecurringOrdersProduct
```
mandan2, line 12: Noticed that multiple repositories use findOneBy and findAllBy methods. These methods could be moved into AbstractRepository and then all repositories could reach them. Simmilar pattern was used in Bolt, Faire I think.

tests/Integration/Subscription/Factory/TestCreateMandateData.php
```diff
@@ -44,7 +44,7 @@ public function testBuild()
         $orderMock = $this->createMock('Order');
         $orderMock->method('getCustomer')->willReturn($customer);
 
-        $CreateMandateData = $mandateDataBuilder->build($orderMock);
+        $CreateMandateData = $mandateDataBuilder->buildFromOrder($orderMock);
```
mandan2, line 47: $createMandateData

subscription/Repository/RecurringOrdersProductRepository.php
margud: the issue is then you cant add return type, it was like that before but because we want to see return type we need to override and use new one with correct return type

mandan2 (approved):
Looks better :)

PR #660, opened 2023-02-06
PIPRES-6: improved subscription grid and page display

margud:
image
image
Improved address and price display


subscription/Validator/CanProductBeAddedToCartValidator.php
```diff
@@ -5,12 +5,13 @@
 namespace Mollie\Subscription\Validator;
 
 use Cart;
```
mandan2, line 7: remove Cart import as adapter is used

subscription/Presenter/RecurringOrderPresenter.php
```diff
@@ -43,15 +42,14 @@ public function present(int $recurringOrderId): array
...
-        $combination = new Combination($recurringProduct->id_product_attribute, null, $this->language->getDefaultLanguageId());
         $order = new Order($recurringOrder->id_order);
         $currency = new Currency($order->id_currency);
 
         $recurringOrderData = [];
         $recurringOrderData['recurring_order'] = $recurringOrder;
         $recurringOrderData['recurring_product'] = $recurringProduct;
         $recurringOrderData['product'] = $product;
-        $recurringOrderData['product_combination_price'] = $product->getPrice(true, $combination->id);
+        $recurringOrderData['total_price'] = $recurringProduct->unit_price * $recurringProduct->quantity;
```
mandan2, line 52: I think problems with this price calculation could happen, like one cent difference. I would suggest "prestashop/decimal" lib for example. 

Other way could be to fetch data from order_detal table and get total_price_tax_incl if your product data is in this table. This is way easier IMO and you wouldn't need to worry about bad calculation.

margud: this is only used for price display, even if there is 1 cent difference it doesn't do much. but yeah i will add Number class still

PR #662, opened 2023-02-07
E2E check #1

SimonasB88:


SimonasB88:
Noticed a possible bug in TestRail huge amount of data transferring, referring https://github.com/boxblinkracer/cypress-testrail/issues/16.

PR #663, opened 2023-02-07
PIPRES-6: improved validations and added atribute adapter

margud:
added attribute adapter and some small improvements

### DIFF
--- a/src/Adapter/ProductAttributeAdapter.php
+++ b/src/Adapter/ProductAttributeAdapter.php
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
 
 namespace Mollie\Adapter;
 
@@ -8,6 +10,7 @@ class ProductAttributeAdapter
 {
     /**
      * @return \AttributeCore|\ProductAttributeCore
+     *
      * @throws MollieException
      */
     public function getProductAttribute(?int $id = null, ?int $idLang = null, ?int $idShop = null)
--- a/src/Adapter/ProductAttributeAdapter.php
+++ b/src/Adapter/ProductAttributeAdapter.php
@@ -1,0 +1,25 @@
+<?php declare(strict_types=1);
+
+namespace Mollie\Adapter;
+
+use Mollie\Exception\MollieException;
+
+class ProductAttributeAdapter
+{
+    /**
+     * @return \AttributeCore|\ProductAttributeCore
+     * @throws MollieException
+     */
+    public function getProductAttribute(?int $id = null, ?int $idLang = null, ?int $idShop = null)
+    {
+        if (class_exists('AttributeCore')) {
+            return new \AttributeCore($id, $idLang, $idShop);
+        }
+
+        if (class_exists('ProductAttributeCore')) {
+            return new \ProductAttributeCore($id, $idLang, $idShop);
+        }
+
+        throw new MollieException('Attribute class was not found');
+    }
+}
--- a/src/Service/MailService.php
+++ b/src/Service/MailService.php
@@ -45,13 +45,13 @@ class MailService
      */
     private $context;
     /** @var ProductAttributeAdapter */
-    private $attributeAdapter;
+    private $productAttributeAdapter;
 
-    public function __construct(Mollie $module, ProductAttributeAdapter $attributeAdapter)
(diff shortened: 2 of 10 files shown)


src/Adapter/ProductAttributeAdapter.php
```diff
@@ -0,0 +1,28 @@
...
+class ProductAttributeAdapter
+{
+    /**
+     * @return \AttributeCore|\ProductAttributeCore
+     *
+     * @throws MollieException
+     */
+    public function getProductAttribute(?int $id = null, ?int $idLang = null, ?int $idShop = null)
+    {
+        if (class_exists('AttributeCore')) {
```
margud, line 18: added core classes because is php8 there is default Attribute class and i m afraid that it might find the php class and not prestashop and that would end in exceptions

src/Service/MailService.php
```diff
@@ -127,7 +130,7 @@ private function getOrderConfData(Order $order, $orderStateId)
 
             $product_price = PS_TAX_EXC == Product::getTaxCalculationMethod() ? Tools::ps_round($price, 2) : $price_wt;
 
-            $attribute = new Attribute($product['product_attribute_id'], $this->context->language->id);
+            $attribute = $this->attributeAdapter->getProductAttribute($product['product_attribute_id'], $this->context->language->id);
```
mandan2, line 133: `$product['product_attribute_id']` I would recommend casting it to int and adding check if it's null (or key does not exist), then pass 0. In this case `getProductAttribute()` first param could be int and not int/null.

subscription/Install/AttributeInstaller.php
```diff
@@ -32,16 +32,21 @@ class AttributeInstaller extends AbstractInstaller
...
+    /** @var ProductAttributeAdapter */
+    private $attributeAdapter;
+
     public function __construct(
         LoggerInterface $logger,
         ConfigurationAdapter $configuration,
         Mollie $module,
-        LanguageRepository $language
+        LanguageRepository $language,
+        ProductAttributeAdapter $attributeAdapter
```
mandan2, line 43: So just call it $productAttributeAdapter, why different naming.

subscription/Install/AttributeInstaller.php
```diff
@@ -97,12 +102,12 @@ private function createAttributeGroup(array $languages): AttributeGroup
     private function createAttributes(array $languages, int $attributeGroupId): void
     {
         foreach (Config::getSubscriptionAttributeOptions() as $attributeName => $attributeConfigKey) {
-            $existingAttribute = new Attribute($this->configuration->get($attributeConfigKey));
+            $existingAttribute = $this->attributeAdapter->getProductAttribute((int) $this->configuration->get($attributeConfigKey));
```
mandan2, line 105: Also you could add check if retrieved value is empty, pass 0.

subscription/Validator/SubscriptionProductValidator.php
```diff
@@ -17,17 +19,21 @@ class SubscriptionProductValidator
     /** @var ProductCombinationRepository */
     private $combinationRepository;
 
-    /** @var \Mollie\Subscription\Repository\CombinationRepository */
+    /** @var CombinationRepository */
     private $combination;
+    /** @var ProductAttributeAdapter */
+    private $attributeAdapter;
```
mandan2, line 25: same, rename it

mandan2 (commented):
Minor issues

src/Service/MailService.php
margud: well this is adapter, and default attribute classes in core has ?int so shouldn't we keep same logic? I guess we can make just int but then we need to be more careful with casting and in core there are log of cases when we get null. So not sure if its worth the risk

subscription/Install/AttributeInstaller.php
margud: what do you mean? if configuration returns null the int cast will change it to 0

src/Service/MailService.php
mandan2: It logic depends on null as you say then it's ok I guess, just leave it as it is. But IMO we use adapter to make methods more usable for us so multiple types could be avoided as much as possible.

subscription/Install/AttributeInstaller.php
mandan2: Yes, correct. Sorry :D

mandan2 (approved):

PR #666, opened 2023-02-14
PIPRES-6/webhook improvments

margud:


src/Service/TransactionService.php
```diff
@@ -138,10 +138,17 @@ public function processTransaction($apiPayment)
         $orderDescription = $apiPayment->description ?? $apiPayment->orderNumber;
         $isGeneratedOrderNumber = strpos($orderDescription, OrderNumberUtility::ORDER_NUMBER_PREFIX) === 0;
         $isPaymentFinished = MollieStatusUtility::isPaymentFinished($apiPayment->status);
+
         if (!$isPaymentFinished && $isGeneratedOrderNumber) {
             return $apiPayment;
         }
 
+        $paymentMethod = $this->paymentMethodRepository->getPaymentBy('transaction_id', $apiPayment->id);
```
mandan2, line 146: Constants could be used instead of strings, way easier to maintain.

subscription/Handler/RecurringOrderCreationHandler.php
```diff
@@ -46,19 +67,51 @@ public function handle(string $transactionId)
         $recurringOrder = $this->recurringOrderRepository->findOneBy(['mollie_subscription_id' => $transaction->subscriptionId]);
         $subscriptionData = $this->subscriptionDataFactory->build((int) $recurringOrder->id);
         $subscription = $this->subscriptionApi->getSubscription($subscriptionData);
-        //todo: add transaction status check
+
+        $key = Mollie\Utility\SecureKeyUtility::generateReturnKey(
```
mandan2, line 71: Why it is not imported?

subscription/Handler/RecurringOrderCreationHandler.php
```diff
@@ -25,19 +34,31 @@ class RecurringOrderCreationHandler
...
         $this->subscriptionDataFactory = $subscriptionDataFactory;
         $this->recurringOrderRepository = $recurringOrderRepository;
         $this->mollie = $mollie;
         $this->mollieOrderCreationService = $mollieOrderCreationService;
+        $this->paymentMethodRepository = $paymentMethodRepository;
+        $this->orderStatusService = $orderStatusService;
+        $this->paymentMethodService = $paymentMethodService;
     }
 
     public function handle(string $transactionId)
```
mandan2, line 64: Add return type

subscription/Handler/RecurringOrderCreationHandler.php
```diff
@@ -46,19 +67,51 @@ public function handle(string $transactionId)
...
+                $this->updateOrderStatus($transaction, (int) $existingTransaction['order_id']);
+                break;
+            case $transaction->status === PaymentStatus::STATUS_PAID:
+                $this->createSubscription($transaction, $recurringOrder, $subscription);
+                break;
+            default:
+                break;
+        }
+
+        return 'OK';
```
mandan2, line 94: Again, constant could be used here. Also, will this 'OK' be returned as response to webhook or something? Why is it not bool for example.

subscription/Handler/RecurringOrderCreationHandler.php
```diff
@@ -46,19 +67,51 @@ public function handle(string $transactionId)
...
+
+        $methodName = $paymentMethod->method_name ?: Config::$methods[$transaction->method];
+
         $newCart = $newCart['cart'];
         $this->mollie->validateOrder(
             (int) $newCart->id,
             Config::getStatuses()[$transaction->status],
             (float) $subscription->amount->value,
-            $transaction->method,
+            'subscription/' . $methodName,
```
mandan2, line 114: Just a tip: sprintf could be used, looks more tidy.

subscription/Presenter/RecurringOrdersPresenter.php
```diff
@@ -43,7 +43,12 @@ public function __construct(
 
     public function present(string $molCustomerId): array
     {
-        $recurringOrders = $this->recurringOrderRepository->findAllBy(['mollie_customer_id' => $molCustomerId]);
+        $recurringOrders = $this->recurringOrderRepository->findAllBy(['mollie_customer_id' => $molCustomerId])->getResults();
```
mandan2, line 46: What if it's missing? getResults would crash

mandan2 (approved):
minor things

src/Service/TransactionService.php
margud: yeah would be good, but now it is used in lot of places and it is like this, so i guess there should be new PR only for this change. Maybe would need to refactor the logic a bit.

subscription/Handler/RecurringOrderCreationHandler.php
margud: no idea to be honest, this return type was used even before we started working on it, and its webhook so i guess anything works here as long as its true. so you can do true or OK but i guess when you are debugging its more easy to understand that its finished. Maybe it could be changed but for now i dont think it matters.

subscription/Presenter/RecurringOrdersPresenter.php
margud: tested, it just returns empty array, all good.

PR #669, opened 2023-02-20
PIPRES-6/added price rule logic to get old price for recurring orders…

margud:
Added faq controller.
Added subscription cancelation from mollie side handle logic
added specific prices for order creation


subscription/Controller/Symfony/SubscriptionFAQController.php
```diff
@@ -0,0 +1,37 @@
...
+    private const FILE_NAME = 'SubscriptionFAQController';
+
+    /**
+     * @AdminSecurity("is_granted('read', request.get('_legacy_controller'))")
+     *
+     * @return Response
+     */
+    public function indexAction()
+    {
+        return $this->render('@Modules/mollie/views/templates/admin/Subscription/subscriptions-faq.html.twig',
```
margud, line 21: Text will be changed by mollie team, now its just the main points from. So dont comment text issues but other part can be reviewed

controllers/admin/AdminMollieSubscriptionFAQController.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+use Mollie\Subscription\Controller\AbstractAdminController;
+
+/**
+ * This controller is only used to create tab and redirect to SubscriptionSettingsController
```
mandan2, line 8: Uhhhh so `AdminMollieSubscriptionFAQParentController` controller has redirect and this one doesn't. Probably forgot to change this here?

subscription/Controller/Symfony/SubscriptionFAQController.php
```diff
@@ -0,0 +1,37 @@
...
+            [
+                'subscriptionCreationTittle' => $this->module->l('Subscription creation', self::FILE_NAME),
+                'subscriptionCreation' => $this->module->l('To create subscription option for product you need to add subscription attribute for the products that you want to offer as subscriptions', self::FILE_NAME),
+                'importantInformationTittle' => $this->module->l('IMPORTANT points', self::FILE_NAME),
+                'importantInformation' => $this->module->l('If you are creating subscription product, make sure you create attribute with option subscription:none to avoid confusing warnings', self::FILE_NAME),
+                'cartRUleTitle' => $this->module->l('Cart rules', self::FILE_NAME),
```
mandan2, line 27: Ru*

subscription/Handler/RecurringOrderHandler.php
```diff
@@ -107,7 +124,13 @@ private function createSubscription(Payment $transaction, MolRecurringOrder $rec
 
         $methodName = $paymentMethod->method_name ?: Config::$methods[$transaction->method];
 
+        /** @var Cart $newCart */
```
mandan2, line 127: Check if 'cart' index actually exists in array and if it's actually Cart.

subscription/Handler/RecurringOrderHandler.php
```diff
@@ -123,11 +146,49 @@ private function createSubscription(Payment $transaction, MolRecurringOrder $rec
...
     }
 
     private function updateOrderStatus(Payment $transaction, int $orderId)
     {
         $this->orderStatusService->setOrderStatus($orderId, $transaction->status);
     }
+
+    private function cancelSubscription(int $recurringOrderId): void
+    {
+        $recurringOrder = new MolRecurringOrder($recurringOrderId);
```
mandan2, line 161: Why you don't use doctrine if other parts are written with symfony?

subscription/Handler/RecurringOrderHandler.php
```diff
@@ -123,11 +146,49 @@ private function createSubscription(Payment $transaction, MolRecurringOrder $rec
...
+        $recurringOrder->date_update = $this->clock->getCurrentDate();
+        $recurringOrder->update();
+    }
+
+    /**
+     * creating temporary specific price for recurring order that will be deleted after order is created
+     */
+    private function createSpecificPrice(MolRecurringOrdersProduct $molRecurringOrdersProduct, MolRecurringOrder $recurringOrder): SpecificPrice
+    {
+        $specificPrice = new SpecificPrice();
```
mandan2, line 174: Is this "specific price" method used in admin orders too? In admin order you can just change product price for specific cart and no data is inserted/deleted additionally for product I think. Not saying your approach is bad or something, but maybe there is a better way.

views/templates/admin/Subscription/subscriptions-faq.html.twig
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+%}
+
+{% block content %}
+  <div class="row">
+    <div class="col-lg-12">
+      <div class="card">
+        <h3 class="card-header">
+          <i class="material-icons">info_outline</i> {{ subscriptionCreationTittle }}
```
mandan2, line 42: why is this string not translated? `info_outline`

views/templates/admin/Subscription/subscriptions-faq.html.twig
```diff
@@ -0,0 +1,120 @@
...
+        <div class="card-block">
+          <div class="card-text">
+            <p>{{recurringOrderCreation}}</p>
+            <p>{{recurringOrderPrice}}</p>
+          </div>
+        </div>
+      </div>
+    </div>
+
+  </div>  <div class="row">
```
mandan2, line 98: lol

mandan2 (changes requested):
Check out if something about specific price could be improved.

views/templates/admin/Subscription/subscriptions-faq.html.twig
margud: this is not string this is icon

subscription/Handler/RecurringOrderHandler.php
margud: it is returned from duplicate function, there is check if success is true, and if its true then it will always be cart here. So i dont think we need to add more validations. If for some reason this is not the case then it would be breaking change from prestashop side.

subscription/Handler/RecurringOrderHandler.php
margud: yes, in admin order page it also create specific price and then deletes it. 
image

mandan2 (approved):

PR #671, opened 2023-02-27
PIPRES-6/added subscription emails for when payments are failed and s…

margud:
image
image
Added emails on subscription payment failed and subscription canceled


src/Service/MailService.php
```diff
@@ -90,20 +111,94 @@ public function sendOrderConfMail(Order $order, $orderStateId)
...
+
+            implode(' ', [$customer->firstname, $customer->lastname]),
             null,
             null,
             $fileAttachment,
             null, _PS_MAIL_DIR_, false, (int) $order->id_shop
         );
     }
 
+    public function sendSubscriptionCancellationWarningMail(int $recurringOrderId): void
```
mandan2, line 130: sendCancelSubscriptionWarningEmail sounds better or sendSubscriptionCancelWarningEmail

src/Service/MailService.php
```diff
@@ -90,20 +111,94 @@ public function sendOrderConfMail(Order $order, $orderStateId)
...
+        $recurringOrderProduct = $this->recurringOrdersProductRepository->findOneBy(['id_mol_recurring_orders_product' => $recurringOrder->id_mol_recurring_orders_product]);
+        $customer = new Customer($recurringOrder->id_customer);
+        $product = new Product($recurringOrderProduct->id_product, false, $customer->id_lang);
+
+        $data = $this->getSubscriptionCancellationWarningMailData($recurringOrder, $recurringOrderProduct, $customer);
+
+        Mail::Send(
+            (int) $customer->id_lang,
+            'mollie_subscription_cancel',
+            sprintf(Mail::l('Your payment for the subscription of %s failed', (int) $customer->id_lang), $product->name),
```
mandan2, line 142: Interesting choice. In one line you use sprintf and in the other one implode.

src/Service/MailService.php
```diff
@@ -90,20 +111,94 @@ public function sendOrderConfMail(Order $order, $orderStateId)
...
+
+        $data = $this->getSubscriptionCancellationWarningMailData($recurringOrder, $recurringOrderProduct, $customer);
+
+        Mail::Send(
+            (int) $customer->id_lang,
+            'mollie_subscription_payment_failed',
+            sprintf(Mail::l('Your subscription for %s cancelled', (int) $customer->id_lang), $product->name),
+            $data,
+            $customer->email,
+            $customer->firstname . ' ' . $customer->lastname,
```
mandan2, line 171: Here you don't use nor sprintf, nor implode

src/Service/MailService.php
```diff
@@ -90,20 +111,94 @@ public function sendOrderConfMail(Order $order, $orderStateId)
...
+        );
+    }
+
+    private function getSubscriptionCancellationWarningMailData(
+        MolRecurringOrder $recurringOrder,
+        MolRecurringOrdersProduct $recurringOrderProduct,
+        Customer $customer
+    ): array {
+        $product = new Product($recurringOrderProduct->id_product, false, $customer->id_lang);
+        $totalPrice = (new Number((string) $recurringOrderProduct->unit_price))->times(new Number((string) $recurringOrderProduct->quantity));
```
mandan2, line 188: Maybe not now but I would strongly recommend to create some sort of getter, which already returns calculated price with this Number lib. It would reduce fault probability and code repeat.

src/Service/MailService.php
```diff
@@ -90,20 +111,94 @@ public function sendOrderConfMail(Order $order, $orderStateId)
...
+            null,
+            null,
+            null,
+            $this->module->getLocalPath() . 'mails/',
+            false,
+            (int) $customer->id_shop
+        );
+    }
+
+    private function getSubscriptionCancellationWarningMailData(
```
mandan2, line 182: DTO would fit here very well. Also it would be more readable if you would use getters instead of array result.

subscription/Handler/RecurringOrderHandler.php
```diff
@@ -156,6 +160,21 @@ private function updateOrderStatus(Payment $transaction, int $orderId)
         $this->orderStatusService->setOrderStatus($orderId, $transaction->status);
     }
 
+    private function handleFailedTransaction(int $recurringOrderId): void
+    {
+        $subscriptionData = $this->subscriptionDataFactory->build($recurringOrderId);
+        $molSubscription = $this->subscriptionApi->getSubscription($subscriptionData);
+        switch ($molSubscription->status) {
```
mandan2, line 167: \n please

mandan2 (changes requested):

src/Service/MailService.php
margud: sprintf is used when i want to add value in middle of string, implode just joins 2 strings

mandan2 (approved):
LOOKS MUCH BETTER GOOD JOB MAN KEEP GOING

PR #673, opened 2023-03-06
PIPRES-6: added search and filter for subscriptions in admin page

margud:
Added subscription search and filter. Fixed filter errors. Also tried to fix automated test but with no luck. for now skipping it and will come back when main feature is done.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,6 @@ jobs:
     strategy:
       matrix:
         php-versions: [ '7.2' ]
-
     steps:
       - name: Checkout
         uses: actions/checkout@v2.0.0
@@ -31,10 +30,10 @@ jobs:
       - run: composer i
 
       - name: Pull PrestaShop files (Tag 1.7.7.0)
-        run: docker run -tid --rm -v ps-volume:/var/www/html -v $PWD/:/var/www/html/modules/mollie --name temp-ps prestashop/prestashop:1.7.7.0
+        run: docker run -tid --rm -v ps-volume:/var/www/html -v $PWD/:/var/www/html/modules/mollie --name temp-ps prestashop/prestashop:1.7.8.1-7.4-apache
 
       - name: Run PHPunit
-        run: docker run --rm --volumes-from temp-ps --workdir=/var/www/html/modules/mollie phpunit/phpunit:5.7.12 -c tests/Unit/
+        run: docker run --rm --volumes-from temp-ps --workdir=/var/www/html/modules/mollie php ./vendor/bin/phpunit -c ./tests/phpunit.xml -c tests/
 
   phpstan:
     name: PHPStan
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,14 +14,13 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: [ '7.2' ]
+
     steps:
       - name: Checkout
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -32,10 +27,11 @@ jobs:
       - run: composer i
 
       - name: Pull PrestaShop files (Tag 1.7.7.0)
-        run: docker run -tid --rm -v ps-volume:/var/www/html -v $PWD/:/var/www/html/modules/mollie --name temp-ps prestashop/prestashop:1.7.7.0
+        run: docker run -tid --rm -v ps-volume:/var/www/html -v $PWD/:/var/www/html/modules/mollie --name temp-ps prestashop/prestashop:1.7.8.1-7.4-apache
```
margud, line 30: ps, tests on github not working right now. tried to fix it but took to much time so added task to fix it after main feature is done.

subscription/Grid/SubscriptionGridQueryBuilder.php
```diff
@@ -100,28 +98,66 @@ private function getQueryBuilder(array $filters): QueryBuilder
...
 
         foreach ($filters as $filterName => $value) {
-            if (!array_key_exists($filterName, $allowedFiltersMap)) {
+            if ('fullname' === $filterName) {
+                $qb->andWhere($likeComparisonFilters[$filterName] . ' LIKE :' . $filterName)
+                    ->setParameter($filterName, '%' . $value . '%');
                 continue;
             }
 
             if ('id_attribute_subgroup' === $filterName) {
```
mandan2, line 125: This key was deleted in array above, so statement won't be true.

subscription/Grid/SubscriptionGridQueryBuilder.php
```diff
@@ -100,28 +98,66 @@ private function getQueryBuilder(array $filters): QueryBuilder
...
                 continue;
             }
 
-            $qb->andWhere($allowedFiltersMap[$filterName] . ' LIKE :' . $filterName)
-                ->setParameter($filterName, '%' . $value . '%')
-            ;
+            if (array_key_exists($filterName, $dateComparisonFilters)) {
+                $alias = $dateComparisonFilters[$filterName];
+
+                foreach ($value as $name => $dateValue) {
```
mandan2, line 140: Have you tested edge cases? E.g. select from date to be same as some record date, same for end date.

subscription/Grid/SubscriptionGridQueryBuilder.php
```diff
@@ -100,28 +98,66 @@ private function getQueryBuilder(array $filters): QueryBuilder
...
+            }
         }
     }
+
+    /**
+     * @return string
+     */
+    private function getNameField(): string
+    {
+        return 'CONCAT(customer.firstname, " ", customer.lastname)';
```
mandan2, line 161: Utility could be used to join two strings

mandan2 (changes requested):

subscription/Grid/SubscriptionGridQueryBuilder.php
margud: yes tested it

subscription/Grid/SubscriptionGridQueryBuilder.php
margud: look again :D this is one string. we add this into mysql to call concat that joins 2 strings using mysql

mandan2 (approved):

PR #675, opened 2023-03-07
PIPRES-6/small bug fixes liek removing not needed buttons and payment…

margud:
… save error with order api payments

### DIFF
--- a/src/Service/MollieOrderCreationService.php
+++ b/src/Service/MollieOrderCreationService.php
@@ -100,7 +100,16 @@ class MollieOrderCreationService
         return $apiPayment;
     }
 
-    public function createMolliePayment(MolliePaymentAlias $apiPayment, $cartId, $orderReference, $orderId = null)
+    /**
+     * @param MolliePaymentAlias|MollieOrderAlias $apiPayment
+     * @param int $cartId
+     * @param string $orderReference
+     * @param int $orderId
+     *
+     * @return void
+     * @throws \PrestaShopDatabaseException
+     */
+    public function createMolliePayment($apiPayment, $cartId, $orderReference, $orderId = null): void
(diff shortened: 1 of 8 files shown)


src/Service/MollieOrderCreationService.php
```diff
@@ -100,7 +100,16 @@ public function createMollieApplePayDirectOrder($paymentData, $paymentMethodObj)
...
+    /**
+     * @param MolliePaymentAlias|MollieOrderAlias $apiPayment
+     * @param int $cartId
+     * @param string $orderReference
+     * @param int $orderId
+     *
+     * @return void
+     * @throws \PrestaShopDatabaseException
+     */
+    public function createMolliePayment($apiPayment, $cartId, $orderReference, $orderId = null): void
```
mandan2, line 112: Why you don't just use types for params?

src/Service/MollieOrderCreationService.php
```diff
@@ -100,7 +100,16 @@ public function createMollieApplePayDirectOrder($paymentData, $paymentMethodObj)
         return $apiPayment;
     }
 
-    public function createMolliePayment(MolliePaymentAlias $apiPayment, $cartId, $orderReference, $orderId = null)
+    /**
+     * @param MolliePaymentAlias|MollieOrderAlias $apiPayment
+     * @param int $cartId
+     * @param string $orderReference
+     * @param int $orderId
```
mandan2, line 107: this one would be int|null then

mandan2 (changes requested):
Noticed that other methods in class doesn't have param types too so it's best that you add them as PHP version supports it. Also, strict_types could be used, then tests would be more accurate and we would catch errors sooner.

PR #677, opened 2023-03-13
fixed PIPRES-192, PIPRES-193, PIPRES-194, PIPRES-195, PIPRES-196, PIP…

margud:
…RES-197, PIPRES-198, PIPRES-201 issues

### DIFF
--- a/mollie.php
+++ b/mollie.php
@@ -386,10 +386,8 @@ class Mollie extends PaymentModule
             }
         }
 
-        $moduleName = Tools::getValue('configure');
-
         // We are on module configuration page
-        if ($this->name === $moduleName && 'AdminMollieSettings' === $currentController) {
+        if ('AdminMollieSettings' === $currentController) {
             $this->context->controller->addJqueryPlugin('sortable');
             $this->context->controller->addJS($this->getPathUri() . 'views/js/admin/payment_methods.js');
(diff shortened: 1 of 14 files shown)


mollie.php
```diff
@@ -386,10 +386,8 @@ public function hookActionAdminControllerSetMedia()
             }
         }
 
-        $moduleName = Tools::getValue('configure');
-
         // We are on module configuration page
-        if ($this->name === $moduleName && 'AdminMollieSettings' === $currentController) {
+        if ('AdminMollieSettings' === $currentController) {
```
mandan2, line 390: Don't hardcode controller name. use AdminMollieSettings::class

subscription/Handler/RecurringOrderHandler.php
```diff
@@ -53,6 +54,8 @@ class RecurringOrderHandler
     private $shop;
     /** @var MailService */
     private $mailService;
+    /** @var TransactionService */
+    private $transactionService;
```
mandan2, line 58: Hm? It's not even used.

mandan2 (approved):
Two minor issues

mollie.php
margud: doesnt work like that for controllers :/ because class is AdminMollieSettingsController but we need to check if its AdminMollieSettings

PR #678, opened 2023-03-13
PIPRES-186: Vendor package PHP8 support

mandan2:


mollie.php
```diff
@@ -107,17 +108,22 @@ private function loadEnv()
             return;
         }
 
-        if (file_exists(_PS_MODULE_DIR_ . 'mollie/.env')) {
-            $dotenv = \Dotenv\Dotenv::create(_PS_MODULE_DIR_ . 'mollie/', '.env');
+        $dovenv = new Dotenv();
```
margud, line 111: why naming is dov and not dot?

mandan2: heh nice typo :D

PR #692, opened 2023-03-28
fixed error handler for ps8 and other small improvments

margud:


mandan2 (approved):
I see in entity you changed some types. Testers probably will need to re-install module to update database. Inform them.

PR #701, opened 2023-04-04
upgrade fixes

margud:


mandan2 (commented):
will need to re-check this tomorrow, maybe will notice some issues in the morning

mandan2 (commented):
in install() method use DI and inject mollie and subscription installers with their interfaces

PR #702, opened 2023-04-04
PIPRES-199: Subscription customer address update

mandan2:
Implemented two hooks in order to update customer addresses for subscription orders. Used trait to have access to property in the same runtime (didn't want to fill mollie.php anymore). How the flow works:

- At first original subscription order has a link with customer address. After we update this address, new address gets created and the old one gets "deleted" flag. We insert newly created address ID into mol_recurring_order table to avoid updating original subscription order data.
- If we try to update newly created address, this address won't get flagged as deleted and will just be updated in the same instance, so our handler won't be triggered. This happens as newly created address was not linked to any of the existing orders ir ps_orders table. Same address ID keeps being used so our flow is not needed.

In mol_recurring_order table created two ned address id columns to save most up to date addresses there, without changing original order data.
Also created decorator for original PrestaShop order presenter lazy array as we wanted to show addresses from mol_recurring_order order, not from original subscription order.

### DIFF
--- a/mollie.php
+++ b/mollie.php
@@ -153,6 +153,7 @@ class Mollie extends PaymentModule
             return false;
         }
 
+//        TODO inject base install and subscription services
         $installer = new \Mollie\Install\Installer(
             $this,
             new \Mollie\Service\OrderStateImageService(),
@@ -164,6 +165,7 @@ class Mollie extends PaymentModule
             ),
             new \Mollie\Adapter\ConfigurationAdapter()
         );
+
         if (!$installer->install()) {
             $this->_errors = array_merge($this->_errors, $installer->getErrors());
 
--- a/mollie.php
+++ b/mollie.php
@@ -33,6 +33,8 @@ require_once __DIR__ . '/vendor/autoload.php';
 
 class Mollie extends PaymentModule
 {
+    use Mollie\Subscription\Traits\HookTraits;
+
     const DISABLE_CACHE = true;
 
     /** @var \Mollie\Api\MollieApiClient|null */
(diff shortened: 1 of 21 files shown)


mollie.php
```diff
@@ -33,6 +33,8 @@
 
 class Mollie extends PaymentModule
 {
+    use Mollie\Subscription\Traits\HookTraits;
```
margud, line 36: import namespace?

subscription/Decorator/RecurringOrderLazyArray.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * @arrayAccess
+     *
+     * @return array
+     */
+    public function getAddresses(): array
+    {
+        $order = $this->order;
+
+        $orderAddresses = [
```
margud, line 49: i guess you could create object here and not array but optional

mandan2: I must use this same pattern as original OrderLazyArray uses array with those specific keys.

subscription/Repository/RecurringOrderRepository.php
```diff
@@ -24,4 +26,23 @@ public function findAllBy(array $keyValueCriteria): ?PrestaShopCollection
 
         return $result;
     }
+
+    public function getCustomerOrdersByOldAddress(int $customerId, int $oldAddressId): ?array
```
margud, line 30: are you using this function?

mandan2: Nice, it is not used, forgot to delete.

margud (approved):

PR #709, opened 2023-04-11
PIPRES-214: Stop add to cart action on subscription product fix

mandan2:
Problem: existing logic for denying product for adding it to the cart (making it available_for_order = false) doesn't work for 1.7.6 version as ps_shoppingcart still shows added to cart modal with no actual added product to cart.

Issue persists only on 1.7.6 version, due to ps_shoppingcart.js file not checking if updateCart emitted action has any errors inside. ps_shoppingcart shows "added to cart" modal, even though operation failed. 
Tested solution on 1.7.6, 1.7.7, 1.7.8

### DIFF
--- a/mollie.php
+++ b/mollie.php
@@ -972,7 +972,6 @@ class Mollie extends PaymentModule
         $orderListActionBuilder = $module->getService(\Mollie\Presenter\OrderListActionBuilder::class);
 
         return $orderListActionBuilder->buildOrderPaymentResendButton($orderId);
-        return $orderListActionBuilder->buildOrderPaymentResendButton($orderId);
     }
 
     public function updateApiKey($shopId = null)
@@ -1018,7 +1017,7 @@ class Mollie extends PaymentModule
 
         http_response_code(Response::HTTP_BAD_REQUEST);
 
-        die(json_encode(
+        exit(json_encode(
             [
                 'hasError' => $hasError,
                 'errors' => $errors,
--- a/mollie.php
+++ b/mollie.php
@@ -349,7 +349,7 @@ class Mollie extends PaymentModule
 
             Media::addJsDef([
                 'mollieSubAjaxUrl' => $this->context->link->getModuleLink('mollie', 'ajax'),
-                'isVersionHigherThan176' => PsVersionUtility::isPsVersionHigherOrEqualsThan(_PS_VERSION_, '1.7.7.0'),
+                'isVersionGreaterThan176' => PsVersionUtility::isPsVersionGreaterOrEqualTo(_PS_VERSION_, '1.7.7.0'),
             ]);
         }
         if (!$paymentMethod || !$paymentMethod->enabled) {
@@ -893,7 +893,7 @@ class Mollie extends PaymentModule
 
     public function hookDisplayProductActions($params)
     {
-        if (PsVersionUtility::isPsVersionHigherOrEqualsThan(_PS_VERSION_, '1.7.6.0')) {
+        if (PsVersionUtility::isPsVersionGreaterOrEqualTo(_PS_VERSION_, '1.7.6.0')) {
             return $this->display(__FILE__, 'views/templates/front/apple_pay_direct.tpl');
         }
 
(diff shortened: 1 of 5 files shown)


mollie.php
```diff
@@ -345,8 +346,10 @@ public function hookActionFrontControllerSetMedia($params)
         if ($isProductController) {
             $this->context->controller->addJS("{$this->_path}views/js/front/subscription/product.js");
             $this->context->controller->addJqueryPlugin('growl');
+
             Media::addJsDef([
                 'mollieSubAjaxUrl' => $this->context->link->getModuleLink('mollie', 'ajax'),
+                'isVersionHigherThan176' => PsVersionUtility::isPsVersionHigherOrEqualsThan(_PS_VERSION_, '1.7.7.0'),
```
tomjas1997, line 352: isPsVersionHigherThan(_PS_VERSION_, '1.7.6.0') :]

views/js/front/subscription/product.js
```diff
@@ -1,15 +1,38 @@
...
+      }
+
+      if (
+        settings.url.toLowerCase().indexOf('controller=cart') > 0 &&
+        settings.data.toLowerCase().indexOf('action=update') > 0 &&
+        settings.data.toLowerCase().indexOf('add=1') > 0
+      ) {
+        validateProduct(getProductData());
+      }
+    });
```
tomas862, line 14: I assume update product event should emit some kind of event you could listen to ?

mollie.php
mandan2: isPsVersionGreaterOrEqualTo fits better imo, it's from bolt as well

views/js/front/subscription/product.js
```diff
@@ -1,15 +1,38 @@
...
+    prestashop.on('updateCart', function() {
+        validateProduct(getProductData());
     });
 
+    function getProductData()
+    {
+      let productDetails = $('#product-details').attr('data-product');
+
+      if (productDetails.length < 1) {
+        return;
```
tomjas1997, line 25: Returns nothing and then passes it to validateProduct, pretty sure there's an error here.

views/js/front/subscription/product.js
```diff
@@ -1,15 +1,38 @@
...
-            {
-                'id_product': productDetails.id_product,
-                'id_product_attribute': productDetails.id_product_attribute,
-            }
+    $(document).ajaxComplete(function (event, xhr, settings) {
+      if (isVersionHigherThan176) {
+        return;
+      }
+
+      if (
```
tomjas1997, line 7: prestashop.on('updateCart', function() {  should really perform all of this validation itself.

mollie.php
mandan2: isVersionLessThan177 maybe??

tomjas1997: Just rename key to 'isVersionGreaterOrEqualTo177' => PsVersionUtility::isPsVersionGreaterOrEqualsThan(_PS_VERSION_, '1.7.7.0'),

views/js/front/subscription/product.js
mandan2: I understand that it should. But the problem is ps_shoppingcart.js file. It also waits for updateCart event and then performs ajax call to retrieve modal. On 1.7.7. version this file has check if event.hasError and then stops, but on 1.7.6 it doesn't exist.
I'm open to hear better solutions as this was best that I could do so far.

`
    prestashop.on(
      'updateCart',
      function (event) {
        var refreshURL = $('.blockcart').data('refresh-url');
        var requestData = {};

        if (event && event.reason) {
          requestData = {
            id_product_attribute: event.reason.idProductAttribute,
            id_product: event.reason.idProduct,
            action: event.reason.linkAction
          };
        }

        $.post(refreshURL, requestData).then(function (resp) {
          $('.blockcart').replaceWith($(resp.preview).find('.blockcart'));
          if (resp.modal) {
            showModal(resp.modal);
          }
        }).fail(function (resp) {
          prestashop.emit('handleError', {eventType: 'updateShoppingCart', resp: resp});
        });
      }
    );
`

views/js/front/subscription/product.js
mandan2: I'll check.

mollie.php
mandan2: ok :c

views/js/front/subscription/product.js
mandan2: nice catch, returned null and checked inside validateProduct method if data is not null

views/js/front/subscription/product.js
mandan2: There is none. updateCart fires after cart is updated, updateProduct fires when variations are changed.

tomjas1997 (approved):
Looks weird, but i guess it works.

PR #719, opened 2023-05-08
PIPRES-202/optional parameters

tomjas1997:
Added optional parameters to be sent:
Title,
Product url,
Product image url,
ConsumerDateOfBirth

### DIFF
--- a/src/Service/CartLinesService.php
+++ b/src/Service/CartLinesService.php
@@ -13,6 +13,7 @@
 namespace Mollie\Service;
 
 use Cart;
+use Mollie\Adapter\Context;
 use Mollie\Adapter\ToolsAdapter;
 use Mollie\Config\Config;
 use Mollie\DTO\Line;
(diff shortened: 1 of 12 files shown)


src/Service/CartLinesService.php
```diff
@@ -194,6 +194,8 @@ private function createProductLines(array $cartItems, $apiRoundingPrecision, $gi
                         'unitPrice' => 0,
                         'totalAmount' => 0,
                         'category' => '',
+                        'product_url' => \Context::getContext()->link->getProductLink($cartItem['id_product']),
+                        'image_url' => \Context::getContext()->link->getImageLink($cartItem['link_rewrite'], $cartItem['id_image']),
```
tomas862, line 198: globals should be avoided in DTO, inject Link via constructor

src/Service/PaymentMethodService.php
```diff
@@ -366,9 +366,20 @@ public function getPaymentData(
             $orderData->setOrderNumber($orderReference);
             $orderData->setLocale($this->getLocale($molPaymentMethod->method));
             $orderData->setEmail($customer->email);
+
+            $gender = new \Gender((int) $customer->id_gender);
```
tomas862, line 370: would prefer to use always repositories, even if you get single entity and it seems that its faster to access object like this

tomas862 (changes requested):

tomas862 (approved):

PR #728, opened 2023-05-17
PIPRES-218: Prevent guest to buy subscription item

mandan2:
also found out that sentry 3.18.0 version causes issues due to missing other library. Downgraded to 3.17.0 and added missing library. Now prestashop doesn't crash.

image

### DIFF
--- a/mollie.php
+++ b/mollie.php
@@ -1181,7 +1181,7 @@ class Mollie extends PaymentModule
 
     public function hookActionFrontControllerAfterInit(): void
     {
-        /**
+        /*
          * NOTE: only logged in customers can buy subscription items
          */
         if (!$this->context->controller instanceof OrderControllerCore) {
--- a/mollie.php
+++ b/mollie.php
@@ -30,7 +30,7 @@ use Mollie\Subscription\Logger\NullLogger;
 use Mollie\Subscription\Repository\LanguageRepository as LanguageAdapter;
 use Mollie\Subscription\Repository\RecurringOrderRepositoryInterface;
 use Mollie\Subscription\Validator\CanProductBeAddedToCartValidator;
-use Mollie\Subscription\Validator\SubscriptionProductValidator;
+use Mollie\Subscription\Verification\HasSubscriptionProductInCart;
 use Mollie\Utility\PsVersionUtility;
 use Symfony\Component\Dotenv\Dotenv;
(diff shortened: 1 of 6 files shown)


mollie.php
```diff
@@ -1177,6 +1179,39 @@ public function hookActionObjectAddressDeleteAfter(array $params): void
...
+    public function hookActionFrontControllerAfterInit(): void
+    {
+        /*
+         * NOTE: only logged in customers can buy subscription items
+         */
+        if (!$this->context->controller instanceof OrderControllerCore) {
+            return;
+        }
+
+        if ($this->context->customer->isLogged()) {
```
tomjas1997, line 1191: Maybe if (!$this->context->customer->isLogged()) { based on the previous NOTE.

subscription/Install/HookInstaller.php
```diff
@@ -39,6 +39,7 @@ private function getHooks(): array
             'actionObjectAddressDeleteAfter',
             'actionBeforeCartUpdateQty',
             'actionAjaxDieCartControllerDisplayAjaxUpdateBefore',
+            'actionFrontControllerAfterInit',
```
tomjas1997, line 42: During upgrade this should be installed as well.

mandan2: in upgrade-6.0.0 file installer service is called so hooks are installed together inside

mollie.php
mandan2: If customer is not logged in, only then we need to run whole logic to check if item is actually subscription item and redirect customer. Otherwise if customer is logged in, we don't need to do anything.

Added additional note message for more context

mollie.php
```diff
@@ -1177,6 +1179,40 @@ public function hookActionObjectAddressDeleteAfter(array $params): void
         $this->addPreventDeleteErrorMessage();
     }
 
+    public function hookActionFrontControllerAfterInit(): void
```
tomjas1997, line 1182: Would really like to have some kind of "Action" for this instead, could extract "searchForSubscriptionProduct" function instead of having notes. 

Leaving everything in main module file has increased it's size drastically (1200 lines.)

mandan2: Moved check if product in cart is subscription to "verification" service.

tomjas1997 (approved):

PR #732, opened 2023-05-23
PIPRES-65: payment method fee usage improvements

mandan2:
Added implementation improvements for paymentFeeProvider usage. Had to refactor some parts as some parts used static method, some used adapter.
Some manual testing is still needed to check if nothing is left out. Wrote one more additional test for paymentFeeProvider, which helped to find some issues.


mandan2:
Create new upgrade file and move changes there as this funcionality won't come with 6.0.0 version

src/Exception/FailedToProvidePaymentFeeException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Mollie\Exception;
+
+class FailedToProvidePaymentFeeException extends \Exception
+{
+    public const FAILED_TO_FIND_TAX_RULES = 1;
```
tomjas1997, line 7: Could start using ExceptionCode class to store all constants

src/Provider/PaymentFeeProvider.php
```diff
@@ -36,23 +36,140 @@
...
+        $this->taxUtility = $taxUtility;
+        $this->taxRuleRepository = $taxRuleRepository;
+        $this->taxRepository = $taxRepository;
     }
 
-    public function getPaymentFee(MolPaymentMethod $paymentMethod)
+    /**
+     * {@inheritDoc}
+     */
+    public function getPaymentFee(MolPaymentMethod $paymentMethod, float $totalCartPrice): PaymentFeeData
```
tomjas1997, line 82: It's a long PR, but could've just passed boolean $taxIncl, to return whichever you want instead of 

`$paymentFeeData = $this->paymentFeeProvider->getPaymentFee($paymentMethod, (float) $originalAmountWithTax);

            $paymentFee = $paymentFeeData->getPaymentFeeTaxIncl();`

each time, not required change.

tests/Unit/Provider/PaymentFeeProviderTest.php
```diff
@@ -0,0 +1,168 @@
...
+        array $expectedResult
+    ): void {
+        $molPaymentMethod = $this->createMock(MolPaymentMethod::class);
+
+        $molPaymentMethod->surcharge = $paymentMethod['surcharge'];
+        $molPaymentMethod->surcharge_percentage = $paymentMethod['surcharge_percentage'];
+        $molPaymentMethod->surcharge_limit = $paymentMethod['surcharge_limit'];
+        $molPaymentMethod->surcharge_fixed_amount_tax_incl = $paymentMethod['surcharge_fixed_amount_tax_incl'];
+        $molPaymentMethod->surcharge_fixed_amount_tax_excl = $paymentMethod['surcharge_fixed_amount_tax_excl'];
+        $molPaymentMethod->tax_rules_group_id = $paymentMethod['tax_rules_group_id'];
```
tomjas1997, line 53: It's always 1, no need to pass it via params.

src/Provider/PaymentFeeProvider.php
```diff
@@ -36,23 +36,140 @@
...
+                break;
+            case Config::FEE_NO_FEE:
+            default:
+                return new PaymentFeeData(0.00, 0.00, true);
+        }
+
+        $surchargeMaxValue = new DecimalNumber((string) $paymentMethod->surcharge_limit);
+        $lowestValue = new DecimalNumber((string) self::LOWEST_VALUE);
+
+        if ($surchargeMaxValue->isGreaterThan($lowestValue) && $totalFeePriceTaxIncl->isGreaterOrEqualThan($surchargeMaxValue)) {
```
tomjas1997, line 159: Lowest value does not really change, it's always 0.

`  !$surchargeMaxValue->isLowerOrEqualThanZero()`

src/Provider/PaymentFeeProvider.php
```diff
@@ -36,23 +36,140 @@
...
+                        $maxPercentage
+                    )
+                );
+
+                $totalFeePriceTaxIncl = $this->taxUtility->addTax(
+                    (float) $totalFeePriceTaxExcl->toPrecision(self::TEMPORARY_PRECISION, Rounding::ROUND_HALF_UP),
+                    $tax
+                );
+
+                $totalFeePriceTaxIncl = new DecimalNumber((string) $totalFeePriceTaxIncl);
```
tomjas1997, line 133: Is there a reason to not return return new PaymentFeeData(...) here?

src/Provider/PaymentFeeProvider.php
```diff
@@ -36,23 +36,140 @@
...
+
+                break;
+            case Config::FEE_NO_FEE:
+            default:
+                return new PaymentFeeData(0.00, 0.00, true);
+        }
+
+        $surchargeMaxValue = new DecimalNumber((string) $paymentMethod->surcharge_limit);
```
tomjas1997, line 156: It seems like it's combining two actions, maybe could use return new PaymentFeeData(...)  for switch cases and handle surchargeMaxValue case in other function? 

Like handleSurcharge()...

src/Provider/PaymentFeeProvider.php
mandan2: Yes, for percentage fee there is a max limit. If percentage fee is bigger, max value will be used.

src/Provider/PaymentFeeProvider.php
mandan2: SurchargeMaxValue is used in percentage and percentage/fixed cases. So only default and fixed fee cases could straight return result. But maybe moving logic to separate method is easier to read, will check it out.

tests/Unit/Provider/PaymentFeeProviderTest.php
mandan2: ok

src/Provider/PaymentFeeProvider.php
mandan2: I see your point. Re-checked PR and I see that I don't really use tax excl price for fee at all. Of course, haven't tested all cases where it's used but from reading code, no part used tax excl. If I will have some time, will think about simplifying it.

src/Exception/FailedToProvidePaymentFeeException.php
mandan2: If I will have time, will do. Adding PR anyway as we will try to increase testable code, exception code usage will increase as well.

PR #735, opened 2023-06-05
PIPRES-217: Upgrade SQL file issue fix

mandan2:
Previous SQL call failed due to several reasons:

- Without specifying table_schema (db name), query tries to check other PrestaShop databases if there are ones.
- execute() returns bool of result. But result was not straight value, it was key->value pair so it always failed. getValue returns first row's value so now it gives us correct result.

### DIFF
--- a/upgrade/Upgrade-6.0.0.php
+++ b/upgrade/Upgrade-6.0.0.php
@@ -27,14 +27,16 @@ function upgrade_module_6_0_0(Mollie $module): bool
     ';
 
     /** only add it if it doesn't exist */
-    if (!(int) Db::getInstance()->getValue($sql)) {
+    if (!Db::getInstance()->getValue($sql)) {
         $sql = '
         ALTER TABLE ' . _DB_PREFIX_ . 'mollie_payments
         ADD `mandate_id` VARCHAR(64);
         ';
 
         try {
-            Db::getInstance()->execute($sql);
+            if (!Db::getInstance()->execute($sql)) {
+                return false;
+            }
         } catch (Exception $e) {
             return false;
         }
@@ -47,7 +49,7 @@ function upgrade_module_6_0_0(Mollie $module): bool
     ';
 
     /** only add it if it doesn't exist */
-    if (!(int) Db::getInstance()->getValue($sql)) {
+    if (!Db::getInstance()->getValue($sql)) {
         $sql = '
         ALTER TABLE ' . _DB_PREFIX_ . 'mol_payment_method
         ADD COLUMN min_amount decimal(20,6) DEFAULT 0,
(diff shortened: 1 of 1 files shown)


upgrade/Upgrade-6.0.0.php
```diff
@@ -21,38 +21,42 @@
...
-    if (!Db::getInstance()->execute($sql)) {
+    if (!(int) Db::getInstance()->getValue($sql)) {
         $sql = '
         ALTER TABLE ' . _DB_PREFIX_ . 'mollie_payments
         ADD `mandate_id` VARCHAR(64);
         ';
 
-        if (!Db::getInstance()->execute($sql)) {
+        try {
+            Db::getInstance()->execute($sql);
```
tomjas1997, line 37: execute can still return false without throwing an exception, should check if it's true/false as well.

upgrade/Upgrade-6.0.0.php
```diff
@@ -21,38 +21,42 @@
...
-    SELECT COUNT(*) AS count
+    SELECT COUNT(*) > 0 AS count
     FROM information_schema.columns
-    WHERE table_name = "' . _DB_PREFIX_ . 'mol_payment_method" AND column_name = \'min_amount\';
+    WHERE TABLE_SCHEMA = "' . _DB_NAME_ . '" AND table_name = "' . _DB_PREFIX_ . 'mol_payment_method" AND column_name = "min_amount";
     ';
 
     /** only add it if it doesn't exist */
-    if (!Db::getInstance()->execute($sql)) {
+    if (!(int) Db::getInstance()->getValue($sql)) {
```
aleksiuno, line 50: Why conversion to the int is needed here?

upgrade/Upgrade-6.0.0.php
tomjas1997: Or use \Db::getInstance()->getMsgError() to get the error.

upgrade/Upgrade-6.0.0.php
```diff
@@ -21,38 +21,42 @@
...
         $sql = '
         ALTER TABLE ' . _DB_PREFIX_ . 'mol_payment_method
         ADD COLUMN min_amount decimal(20,6) DEFAULT 0,
         ADD COLUMN max_amount decimal(20,6) DEFAULT 0;
         ';
 
-        if (!Db::getInstance()->execute($sql)) {
+        try {
+            Db::getInstance()->execute($sql);
```
aleksiuno, line 58: I suppose it can also return false which would mean sql execute failed and currently it returns `void` which means that everything went OK

aleksiuno: In other words I think you should check what SQL execution returns, if it returns `false`, this method should return `false`

upgrade/Upgrade-6.0.0.php
aleksiuno: If it's really needed, wouldn't it work with `bool` conversion? It would make more sense

tomjas1997 (approved):

PR #738, opened 2023-06-05
PIPRES-217: Duplicating menu tabs on upgrade fix

mandan2:
After upgrade old controller tabs stayed in the database, causing duplicate menu items. Manually removing it before registering new tabs.
On 1.7.6. authorization_roles table was causing issues as on this specific presta version there is no check before if such entries exist. Inserting on top of it caused exceptions.

### DIFF
--- a/upgrade/Upgrade-6.0.0.php
+++ b/upgrade/Upgrade-6.0.0.php
@@ -20,6 +20,155 @@ if (!defined('_PS_VERSION_')) {
 
 function upgrade_module_6_0_0(Mollie $module): bool
 {
+    if (!modifyExistingTables()) {
+        return false;
+    }
+
+    if (!deleteOldTabs($module)) {
+        return false;
+    }
+
+    // only for 1.7.6 version
+    if (version_compare(_PS_VERSION_, '1.7.7.0', '<')) {
+        if (!deleteOldTabAuthorizationRoles($module)) {
+            return false;
+        }
+    }
+
+    //todo: maybe move to container
+    $installer = new \Mollie\Install\Installer(
+        $module,
+        new \Mollie\Service\OrderStateImageService(),
+        new \Mollie\Install\DatabaseTableInstaller(),
+        new \Mollie\Tracker\Segment(
+            new \Mollie\Adapter\Shop(),
+            new \Mollie\Adapter\Language(),
+            new \Mollie\Config\Env()
+        ),
+        new \Mollie\Adapter\ConfigurationAdapter()
+    );
+
+    $installer->installSpecificTabs();
+
+    /** @var ModuleTabRegister $tabRegister */
+    $tabRegister = $module->getService('prestashop.adapter.module.tab.register');
+
+    $moduleAdapter = new \PrestaShop\PrestaShop\Adapter\Module\Module();
+    $moduleAdapter->instance = $module;
+    $moduleAdapter->disk = new ParameterBag(
+        [
+            'filemtype' => 0,
+            'is_present' => 1,
+            'is_valid' => 1,
+            'version' => null,
+            'path' => '',
+        ]
+    );
+
+    $moduleAdapter->attributes->set('name', $module->name);
+
+    $tabRegister->registerTabs($moduleAdapter);
+
+    /** @var InstallerInterface $installer */
+    $installer = $module->getService(InstallerInterface::class);
+
+    return $installer->install();
+}
+
+function deleteOldTabs($module): bool
+{
+    $classNames = [$module::ADMIN_MOLLIE_CONTROLLER, $module::ADMIN_MOLLIE_AJAX_CONTROLLER];
+    $preparedClassNames = [];
+
+    foreach ($classNames as $className) {
+        $preparedClassNames[] = pSQL($className);
+    }
+
+    $preparedClassNames = implode("', '", $preparedClassNames);
+
+    $sql = '
+        SELECT id_tab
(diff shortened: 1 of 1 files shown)


upgrade/Upgrade-6.0.0.php
```diff
@@ -65,6 +65,44 @@ function upgrade_module_6_0_0(Mollie $module): bool
         }
     }
 
+    $sql = '
+    DELETE t, tl
+    FROM `' . _DB_PREFIX_ . 'tab` t
+    JOIN `' . _DB_PREFIX_ . 'tab_lang` tl ON t.id_tab = tl.id_tab
```
tomjas1997, line 71: I don't like joining on DELETE action, you can leave ps_tab_lang table without deleting it, as it doesn't really affect much. Or at at least gather all tabs ID's and then delete based on them in tab_lang table.

upgrade/Upgrade-6.0.0.php
```diff
@@ -65,6 +65,44 @@ function upgrade_module_6_0_0(Mollie $module): bool
...
+    WHERE t.class_name IN (\'' . $module::ADMIN_MOLLIE_CONTROLLER . '\', \'' . $module::ADMIN_MOLLIE_AJAX_CONTROLLER . '\');
```
tomjas1997, line 72: $classNames = [...];
$preparedClassNames = [];

foreach ($classNames as $className) {
 $preparedClassNames[] = '"' . pSQL($className) . '"';
}

Your current functionality may cause errors.

upgrade/Upgrade-6.0.0.php
```diff
@@ -104,3 +74,145 @@ function upgrade_module_6_0_0(Mollie $module): bool
...
     return $installer->install();
 }
+
+function deleteOldTabs($module): bool
+{
+    $classNames = [$module::ADMIN_MOLLIE_CONTROLLER, $module::ADMIN_MOLLIE_AJAX_CONTROLLER];
+    $preparedClassNames = [];
+
+    foreach ($classNames as $className) {
+        $preparedClassNames[] = pSQL($className);
```
tomjas1997, line 84: Each should be surrounded by quotes. Like in the example i gave, currently it looks like "test, test"

upgrade/Upgrade-6.0.0.php
```diff
@@ -104,3 +74,145 @@ function upgrade_module_6_0_0(Mollie $module): bool
...
+    }
+
+    if (empty($tabIds)) {
+        return false;
+    }
+
+    $preparedTabIds = [];
+
+    foreach ($tabIds as $tabId) {
+        $preparedTabIds[] = $tabId['id_tab'];
```
tomjas1997, line 108: Int cast

upgrade/Upgrade-6.0.0.php
```diff
@@ -104,3 +74,145 @@ function upgrade_module_6_0_0(Mollie $module): bool
...
+    }
+
+    $preparedTabIds = implode("', '", $preparedTabIds);
+
+    $sql = '
+    DELETE FROM `' . _DB_PREFIX_ . 'tab`
+    WHERE id_tab IN (\'' . $preparedTabIds . '\');
```
tomjas1997, line 115: Not sure why the single quotes are needed, it's integer array. IN (1,2,3) works perfectly well.

upgrade/Upgrade-6.0.0.php
```diff
@@ -104,3 +74,145 @@ function upgrade_module_6_0_0(Mollie $module): bool
...
+
+    return true;
+}
+
+function deleteOldTabAuthorizationRoles($module): bool
+{
+    $crudActions = ['_CREATE', '_READ', '_UPDATE', '_DELETE'];
+    $preparedTabs = [];
+
+    foreach ($crudActions as $crudAction) {
```
tomjas1997, line 147: Would expect it other way around. You foreach tabs and add crud actions onto them. Tab + _Create , Tab + _Delete, seems way more readable

upgrade/Upgrade-6.0.0.php
mandan2: Each element is actually surrounded by quotes later. I'm using implode to cast array to string and adding ' , ' in between strings. Later I add ' $elements ' quotes before and after. Result is the same.

upgrade/Upgrade-6.0.0.php
```diff
@@ -104,3 +74,145 @@ function upgrade_module_6_0_0(Mollie $module): bool
...
+        WHERE class_name IN (\'' . $preparedClassNames . '\');
+    ';
+
+    try {
+        $tabIds = Db::getInstance()->executeS($sql);
+    } catch (Exception $e) {
+        return false;
+    }
+
+    if (empty($tabIds)) {
```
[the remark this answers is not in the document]
mandan2: true, added return true instead.

upgrade/Upgrade-6.0.0.php
mandan2: ok

upgrade/Upgrade-6.0.0.php
mandan2: okay

upgrade/Upgrade-6.0.0.php
mandan2: ok

tomjas1997 (approved):
I guess it works, unsure why you decided to use \\' instead of just " for imploding and in the query though.

PR #742, opened 2023-06-13
PIPRES-180: Order status soft delete on uninstall

mandan2:


src/Factory/ModuleFactory.php
```diff
@@ -57,4 +57,15 @@ public function getModule(): ?Mollie
 
         return $module;
     }
+
+    public function getModuleName(): ?string
```
tomjas1997, line 61: Not sure if checking for null module really should happen. If module is null a lot more places will crash 🙃 overall would just get module via constructor and get name that way

tomjas1997: Also name would never be empty

src/ServiceProvider/BaseServiceProvider.php
```diff
@@ -106,7 +106,8 @@ public function register(Container $container)
         $this->addService($container, MolCustomerRepository::class, MolCustomerRepository::class)
             ->withArgument('MolCustomer');
 
-        $this->addService($container, UninstallerInterface::class, Mollie\Install\DatabaseTableUninstaller::class);
+        $this->addService($container, UninstallerInterface::class, Mollie\Install\DatabaseTableUninstaller::class)
+            ->withArgument($container->get(ModuleFactory::class));
```
tomjas1997, line 110: Doesn't require any arguments here, it should work by auto-wiring

src/Factory/ModuleFactory.php
mandan2: It doesn't hurt to keep this check as ide cries about return types if I don't check. :D  Anyway, this service won't be reached if module wouldn't exist.

src/ServiceProvider/BaseServiceProvider.php
mandan2: Fixed

PR #744, opened 2023-06-14
PIPRES-65: tax append to cart summary improvements

mandan2:
Improved ajax call to provide order summary block with updated tax price as well. Previously, only total price was updated.
Also improved overall functionality. Instead of relying on given payment fee from front-end, trying to fetch it fresh from the database.

Payment fee prices are shown with tax incl. Tax is 21%. So 12.10 is 10 without tax.

image
image


tomas862 (approved):
looks good in general :)

PR #745, opened 2023-06-14
PIPRES-191: Misleading exception throw on shipment verification

mandan2:
Use case:

When order is created and initial order state is given for order, updateOrderStatus hook is called. On this hook we try to send shipment. CanSendShipment checks if all requirements are met. But on initial order status update requirements won't be met as orderId is still not returned from validateOrder method, so it is not inserted inside database with mollie transaction. Merchants were annoyed with logged warnings and errors every time order was made from this exact verification. 

That's why we created additional verify service, which is used outside canSendShipment. We use this service to verify if payment information is available by trying to find it by order id.

### DIFF
--- a/src/Verification/IsPaymentInformationAvailable.php
+++ b/src/Verification/IsPaymentInformationAvailable.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Mollie\Verification;
+
+use Mollie\Repository\PaymentMethodRepositoryInterface;
+
+class IsPaymentInformationAvailable
+{
+    /** @var PaymentMethodRepositoryInterface */
+    private $paymentMethodRepository;
+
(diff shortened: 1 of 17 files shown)


src/Verification/IsPaymentInformationAvailable.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $payment = $this->paymentMethodRepository->getPaymentBy('order_id', (int) $orderId);
+
+        if (empty($payment)) {
+            return false;
+        }
+
+        if (empty($payment['transaction_id'])) {
+            return false;
+        }
```
tomas862, line 32: can combine to one liner

tomas862 (approved):
small one

mant02 (approved):
approved

PR #750, opened 2023-06-20
PIPRES-266: Order fee refactor to order payment fee

mandan2:


mant02 (approved):

mollie.php
```diff
@@ -719,24 +729,37 @@ public function hookActionEmailSendBefore($params)
...
-            return;
+            return '';
         }
+
         if (!$params['object'] instanceof OrderInvoice) {
-            return;
+            return '';
         }
 
+        $localeRepo = $this->get('prestashop.core.localization.locale.repository');
```
tomas862, line 742: you sure it is available in all your supported presta versions ?

upgrade/Upgrade-6.0.1.php
```diff
@@ -53,5 +62,60 @@ function upgrade_module_6_0_1(Mollie $module): bool
...
+
+        $sql = 'RENAME TABLE ' . _DB_PREFIX_ . 'mol_order_fee TO ' . _DB_PREFIX_ . 'mol_order_payment_fee';
+
+        try {
+            if (!Db::getInstance()->execute($sql)) {
+                return false;
+            }
+        } catch (Exception $e) {
+            return false;
+        }
```
tomas862, line 117: silent ignoring exceptions is very bad practice, you should atleast use logger to log an error what happened

mollie.php
mandan2: In 1.7.6 it's available, I have tested it. Since this version (6.0+) is only for 1.7.6+ versions, it's alright.
In separate fork we will do functionality migration, there I will use old legacy method for formatting price since PrestaShop versions will be fixed there (1.7.0 - 1.7.5).

upgrade/Upgrade-6.0.1.php
mandan2: Added message logging, didn't think of this. :D

tomas862 (approved):

PR #753, opened 2023-06-27
PIPRES-65: Conditional number of DecimalNumber usage

mandan2:
PrestaShop 1.7.6. uses older version of Number vendor package. Added condition to use available class based if it's extended by another.

### DIFF
--- a/src/Utility/NumberUtility.php
+++ b/src/Utility/NumberUtility.php
@@ -12,10 +12,27 @@
 
 namespace Mollie\Utility;
 
+use PrestaShop\Decimal\DecimalNumber;
 use PrestaShop\Decimal\Number;
 
 class NumberUtility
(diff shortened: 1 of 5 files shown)


src/Utility/NumberUtility.php
tomas862, line 1: you can check bolt or square module how we did, each method is used via our abstraction - that way you can be much safer for later on to check method differences etc... Im not sure how big change that would be now

mandan2: ATM I would leave current implementation as it works fine. Added TODO for this as some services use NumberUtility abstractions, others use straight Number/DecimalNumber methods.
Already added comment in jira regarding this functionality improve.

tomas862 (approved):

PR #754, opened 2023-06-27
PIPRES-65: Tax rule select for percentage fee

mandan2:
Improved fee calculation logic: moved logic to separate calculation service, which utilizes NumberUtility service for calculus actions. 
Additionally, fixed minor bugs for price rounding, variable naming.

P.S. Tests takes up half of written code, sorry. :D

### DIFF
--- a/src/Calculator/PaymentFeeCalculator.php
+++ b/src/Calculator/PaymentFeeCalculator.php
@@ -26,7 +26,7 @@ class PaymentFeeCalculator
     {
         $totalFeePriceTaxIncl = $this->taxCalculator->addTaxes($totalFeePriceTaxExcl);
 
-        return $this->returnFormattedResult(
+        return $this->buildPaymentFee(
             $totalFeePriceTaxIncl,
             $totalFeePriceTaxExcl
         );
@@ -51,7 +51,7 @@ class PaymentFeeCalculator
 
         $totalFeePriceTaxExcl = $this->taxCalculator->removeTaxes($totalFeePriceTaxIncl);
 
-        return $this->returnFormattedResult(
+        return $this->buildPaymentFee(
             $totalFeePriceTaxIncl,
             $totalFeePriceTaxExcl
         );
@@ -79,18 +79,18 @@ class PaymentFeeCalculator
 
         $totalFeePriceTaxExcl = $this->taxCalculator->removeTaxes($totalFeePriceTaxIncl);
 
-        return $this->returnFormattedResult(
+        return $this->buildPaymentFee(
             $totalFeePriceTaxIncl,
             $totalFeePriceTaxExcl
         );
     }
 
-    public function calculateSurchargeMaxValue(float $surchargeMaxValue): PaymentFeeData
+    private function calculateSurchargeMaxValue(float $surchargeMaxValue): PaymentFeeData
     {
         $totalFeePriceTaxIncl = $surchargeMaxValue;
         $totalFeePriceTaxExcl = $this->taxCalculator->removeTaxes($totalFeePriceTaxIncl);
 
-        return $this->returnFormattedResult(
+        return $this->buildPaymentFee(
             $totalFeePriceTaxIncl,
             $totalFeePriceTaxExcl
         );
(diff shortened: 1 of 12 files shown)


src/Calculator/PaymentFeeCalculator.php
tomas862, line 1: as mentioned during our call, remember if you expose multiple public functions , are they doing same stuff ? If not, then you are violating SOLID principe S - single responsability. Although the code here is related but its easy to put code under interface and split it

src/Utility/NumberUtility.php
```diff
@@ -33,6 +37,20 @@ public static function getNumber(float $number)
         return new Number((string) $number);
     }
 
+    public static function setDecimalPrecision(
```
tomas862, line 40: hmm, setter ? toPrecision would sound better IMO

src/Utility/NumberUtility.php
```diff
@@ -33,6 +37,20 @@ public static function getNumber(float $number)
...
+        float $number,
+        int $precision,
+        string $roundingMode = self::ROUNDING
+    ): float {
+        if (is_subclass_of(Number::class, DecimalNumber::class)) {
+            $decimalNumber = new DecimalNumber((string) $number);
+        } else {
+            $decimalNumber = new Number((string) $number);
+        }
```
tomas862, line 49: you can create private static method and hide this logic under need. Then everywhere you use this Number just use
```
self::getNumber($number);
```

tomas862: I see you have self::toObject - why not using it ?

tomas862 (changes requested):

src/Service/ApiService.php
```diff
@@ -410,4 +424,29 @@ private function getSurchargeFixedAmountTaxInclPrice(float $priceTaxExcl, int $t
         return (float) (NumberUtility::getNumber($taxCalculator->addTaxes($priceTaxExcl)))
             ->toPrecision($this->context->getComputingPrecision(), Rounding::ROUND_HALF_UP);
     }
+
+    private function setPrecisionForDecimalNumbers(MolPaymentMethod $paymentMethod): MolPaymentMethod
```
tomjas1997, line 428: appendPaymentMethodFees(..)

src/Utility/NumberUtility.php
```diff
@@ -14,9 +14,13 @@
 
 use PrestaShop\Decimal\DecimalNumber;
 use PrestaShop\Decimal\Number;
+use PrestaShop\Decimal\Operation\Rounding;
 
 class NumberUtility
 {
+    private const PRECISION = 6;
```
tomjas1997, line 21: DECIMAL_PRECISION = 2.
FLOAT_PRECISION = 6;

src/Utility/NumberUtility.php
```diff
@@ -33,6 +37,20 @@ public static function getNumber(float $number)
         return new Number((string) $number);
     }
 
+    public static function setDecimalPrecision(
+        float $number,
+        int $precision,
```
tomjas1997, line 42: use int $precision = self::DECIMAL_PRECISION and just use toPrecision function name

src/Service/ApiService.php
mandan2: It is not the case for the method. I need to set precision for retrieved data from DB as in database it has 6 numbers. Renamed toPrecision... as method name in utility changed as well.

src/Calculator/PaymentFeeCalculator.php
mandan2: I have added TODO for this, right now it's fine as it is. In my opinion even better idea would be to have abstract class instead of interface. I have two common private methods, which would be used in all concrete classes.

src/Utility/NumberUtility.php
mandan2: Renamed

src/Utility/NumberUtility.php
mandan2: Removed this method entirely as it is not used anymore.

src/Calculator/PaymentFeeCalculator.php
```diff
@@ -0,0 +1,121 @@
...
+        float $surchargeLimit
+    ): bool {
+        if (NumberUtility::isGreaterThan($totalFeePriceTaxIncl, $surchargeLimit)) {
+            return true;
+        }
+
+        return false;
+    }
+
+    private function returnFormattedResult(
```
tomjas1997, line 110: buildPaymentFee maybe, it's not really readable to have return returnFormattedResult..

src/Calculator/PaymentFeeCalculator.php
```diff
@@ -0,0 +1,121 @@
...
+        float $totalFeePriceTaxIncl,
+        float $totalFeePriceTaxExcl
+    ): PaymentFeeData {
+        return new PaymentFeeData(
+            NumberUtility::toPrecision($totalFeePriceTaxIncl, $this->context->getComputingPrecision()),
+            NumberUtility::toPrecision($totalFeePriceTaxExcl, $this->context->getComputingPrecision()),
+            $this->taxCalculator->getTotalRate(),
+            $totalFeePriceTaxIncl > 0 && $totalFeePriceTaxExcl > 0
```
tomjas1997, line 118: could extract this to a variable for readability, as i'm not sure why this is even needed.

PR #757, opened 2023-07-10
PIPRES-180: Separate order state installer from main install service

mandan2:


src/Install/Uninstall.php
```diff
@@ -94,27 +94,7 @@ private function deleteConfig()
...
-            Config::MOLLIE_STATUS_EXPIRED,
-            Config::MOLLIE_STATUS_REFUNDED,
-            Config::MOLLIE_STATUS_SHIPPING,
-            Config::MOLLIE_STATUS_DEFAULT,
-        ];
-
-        $this->deleteConfigurations(
-            array_merge($configurations, $orderStateConfigurations)
-        );
+        $this->deleteConfigurations($configurations);
```
tomjas1997, line 97: There's no need to remove order states? I don't see uninstaller for them

tomjas1997 (approved):
One comment, otherwise seems OK

PR #766, opened 2023-07-24
PIPRES-260: Add different identifier for initial subscription order

mandan2:
For initial subscription order add different identifier to enable filter data more easily.

### DIFF
--- a/mollie.php
+++ b/mollie.php
@@ -15,6 +15,7 @@ use Mollie\Adapter\Link;
 use Mollie\Adapter\ProductAttributeAdapter;
 use Mollie\Adapter\ToolsAdapter;
 use Mollie\Api\Exceptions\ApiException;
+use Mollie\Api\MollieApiClient;
 use Mollie\Config\Config;
 use Mollie\Exception\ShipmentCannotBeSentException;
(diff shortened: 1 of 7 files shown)


mollie.php
```diff
@@ -111,10 +112,10 @@ public function getService(string $serviceName)
         return $this->containerProvider->getService($serviceName);
     }
 
-    public function getApiClient()
+    public function getApiClient(int $shopId = null, bool $subscriptionOrder = false): ?MollieApiClient
     {
         if (!$this->api) {
-            $this->setApiKey();
```
tomjas1997, line 117: GenerateApiKey, BuildApiKey maybe, not really setting it

mollie.php
```diff
@@ -111,10 +112,10 @@ public function getService(string $serviceName)
         return $this->containerProvider->getService($serviceName);
     }
 
-    public function getApiClient()
```
tomjas1997, line 114: IsSubscriptionEnabled maybe? Seems weird logic overall, would really need to extract it to a service as currently it's very complicated for no real reaosn

mandan2: I have plans to reconstruct whole api builder so this will do at the moment.

tomjas1997 (approved):

PR #769, opened 2023-07-25
PIPRES-208: Payment fee update actions after created order

mandan2:
Moved DB actions as they will need to be reused in separate services.


src/Action/UpdateOrderTotalsWithPaymentFeeAction.php
```diff
@@ -0,0 +1,55 @@
...
+namespace Mollie\Action;
+
+use Exception;
+use Mollie\DTO\UpdateOrderTotalsWithPaymentFeeData;
+use Mollie\Exception\CouldNotUpdateOrderTotalsWithPaymentFee;
+use Mollie\Repository\OrderRepositoryInterface;
+use Mollie\Utility\NumberUtility;
+use Order;
+
+class UpdateOrderTotalsWithPaymentFeeAction
```
tomjas1997, line 12: Would it ever be without Payment Fee?

src/Action/UpdateOrderTotalsWithPaymentFeeAction.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public function run(UpdateOrderTotalsWithPaymentFeeData $updateOrderTotalsWithPaymentFeeData): void
+    {
+        try {
+            /** @var Order $order */
+            $order = $this->orderRepository->findOneBy([
+                'id_order' => $updateOrderTotalsWithPaymentFeeData->getOrderId(),
+            ]);
+
+            $order->total_paid_tax_excl = NumberUtility::toPrecision(
```
tomjas1997, line 33: Looks a bit weird that getOriginalCartAmountTaxIncl would differ to getTransactionAmount, can it ever happen?

src/Action/UpdateOrderTotalsWithPaymentFeeAction.php
mandan2: Not really, removed payment fee title

src/Action/UpdateOrderTotalsWithPaymentFeeAction.php
mandan2: Yes it can. When working with payment fee, we only modify order. Cart is left untouched so it does not have payment fee added.

PR #774, opened 2023-08-01
PIPRES-303: Locale retrieve fix for email template

mandan2:
Context doesn't always provide us with current locale so we need to retrieve it via Symfony's DI service. It would be a waste to create our locale repository to do this as quite some dependencies are needed there in repository.


tomjas1997 (approved):
LGTM.

PR #783, opened 2023-08-21
PIPRES-113: Billie payment method validation

mandan2:
Added validation for Billie payment method. Payment method requirements:
- B2B enabled
- Vat number entered
- Identification number entered

Additionally bumped Mollie vendor package version as Billie was introduced later as we will use constant from the package for saving it in our configs.


src/Service/PaymentMethod/PaymentMethodRestrictionValidation/B2bPaymentMethodRestrictionValidator.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        return true;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function supports(MolPaymentMethod $paymentMethod): bool
+    {
```
tomjas1997, line 59: Should just put all of this logic inside this function tbh

tomjas1997 (approved):

PR #788, opened 2023-08-22
PIPRES-113: Billie payment method setup

mandan2:
As Billie payment method will have same actions as Klarna, we needed to use general naming for such methods. Renamed every const, variable, changed order status names and configuration values.


tomjas1997 (approved):
Seems good, a bit weird though :D

PR #790, opened 2023-08-22
PIPRES-113: Billie payment method updated translations

mandan2:
TODO set all missing EN translations and check if other languages have all the keys.


mandan2:
Update all translations when we will have them in all languages.

PR #796, opened 2023-08-28
PIPRES-261: Shipping option module setting

mandan2:


mandan2:
Currently we chose to use same shipping option for subscription order. Also, shipping price won't be included into subscription API, product price will bundle both prices together.

PR #808, opened 2023-09-12
PIPRES-306: Configuration updateValue multishop context fix

mandan2:


src/Adapter/ConfigurationAdapter.php
```diff
@@ -12,59 +12,51 @@
...
+    public function updateValue($key, $value, $idShop = null, $html = false, $idShopGroup = null): void
     {
-        if (is_array($key)) {
-            if ((int) $this->get(Config::MOLLIE_ENVIRONMENT)) {
-                $key = $key['production'];
-            } else {
-                $key = $key['sandbox'];
-            }
-        }
+        $key = $this->parseKeyByEnvironment($key);
```
tomjas1997, line 52: should use same logic as in our other modules

PR #812, opened 2023-09-18
PIPRES-338: Mixed cart subscription listing improvements

mandan2:
Added alter table improvements for install process as following tables are not being removed during uninstall process at all so there were several merchants, which had previously installed Mollie module and came back to install it after some time. Due to this flow some columns were not appended as no upgrade script was executed.

Improvements for subscription order listing in BO and FO.

In BO instead of unit_price started showing total order price. Additionally changed email data for total price.
In FO improved subscriptions order listing to show only subscription order total price (previously full mixed order price was shown)
In FO added decorated OrderLazyArray class and changed OrderProvider to give only products and totals for subscription context. Adding example of recurring order, which had in the cart only subscription product to show that prices are the same. Original order had multiple random products but only one subscription.

image
Original order:
image
image
Subscription order view without other products
image
New recurring order from webhook
image

### DIFF
--- a/subscription/Presenter/RecurringOrderLazyArray.php
+++ b/subscription/Presenter/RecurringOrderLazyArray.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Mollie\Subscription\Presenter;
+
+use PrestaShop\PrestaShop\Adapter\Presenter\Order\OrderLazyArray;
+
+class RecurringOrderLazyArray extends OrderLazyArray
+{
+    /** @var int */
+    private $recurringOrderProductAttributeId;
+
+    public function setRecurringOrderProductAttributeId(int $recurringOrderProductAttributeId): void
+    {
+        $this->recurringOrderProductAttributeId = $recurringOrderProductAttributeId;
+    }
+
+    /**
+     * @arrayAccess
+     *
+     * @return array
+     */
+    public function getProducts(): array
+    {
+        $subscriptionProduct = [];
+
+        $orderProducts = parent::getProducts();
+
+        foreach ($orderProducts as $orderProduct) {
+            if ((int) $orderProduct['id_product_attribute'] !== $this->recurringOrderProductAttributeId) {
+                continue;
+            }
+
+            $subscriptionProduct[] = $orderProduct;
+
+            break;
+        }
+
+        return $subscriptionProduct;
+    }
+}
--- a/subscription/Presenter/RecurringOrderPresenter.php
+++ b/subscription/Presenter/RecurringOrderPresenter.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Mollie\Subscription\Logger;
+namespace Mollie\Subscription\Presenter;
 
 use Currency;
 use Mollie\Adapter\Language;
@@ -10,7 +10,6 @@ use Mollie\Subscription\Api\MethodApi;
 use Mollie\Subscription\Repository\RecurringOrderRepositoryInterface;
 use Mollie\Subscription\Repository\RecurringOrdersProductRepositoryInterface;
 use Order;
-use PrestaShop\PrestaShop\Adapter\Presenter\Order\OrderPresenter;
 use Product;
 
 class RecurringOrderPresenter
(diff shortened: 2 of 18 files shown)


subscription/Presenter/RecurringOrderLazyArray.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace Mollie\Subscription\Presenter;
+
+use PrestaShop\PrestaShop\Adapter\Presenter\Order\OrderLazyArray;
+
+class RecurringOrderLazyArray extends OrderLazyArray
```
tomjas1997, line 7: Should it really extend?

subscription/Presenter/RecurringOrderPresenter.php
```diff
@@ -56,7 +60,11 @@ public function present(int $recurringOrderId): array
         $recurringOrderData['recurring_order'] = $recurringOrder;
         $recurringOrderData['recurring_product'] = $recurringProduct;
         $recurringOrderData['product'] = $product;
-        $recurringOrderData['order'] = (new OrderPresenter())->present($order);
+        $recurringOrderData['order'] = $this->orderPresenter->present(
+            $order,
+            (int) $recurringProduct->id_product_attribute,
+            NumberUtility::toPrecision((float) $recurringOrder->total_tax_incl, 2)
```
tomjas1997, line 66: constant intead of number

upgrade/Upgrade-6.0.4.php
```diff
@@ -105,3 +109,46 @@ function updateOrderStatusNames604(Mollie $module)
     $authorizablePaymentStatusAuthorized->save();
 }
 
+function modifyExistingTables604(): bool
```
tomjas1997, line 112: did you miss 'alter' => [
                    '
                        ALTER TABLE ' . _DB_PREFIX_ . 'mollie_payments
                        ADD COLUMN mandate_id VARCHAR(64);
                    ',
                ],?

mandan2: No, this column was appended back in 6.0.0 release.

subscription/Presenter/RecurringOrderLazyArray.php
mandan2: We're using theme's existing template `customer/_partials/order-detail-no-return.tpl` and OrderLazyArray for this template is needed. Extension makes sens as we don't need to change other methods, only the ones we need to manipulate.

image

upgrade/Upgrade-6.0.4.php
mandan2: Added this column anyway as we currently have a few tickets for this issue so upgrade should fix it.

tests/Integration/Subscription/Presenter/OrderPresenterTest.php
```diff
@@ -0,0 +1,74 @@
...
+            ],
+        ];
+
+        $order = $this->createMock(\Order::class);
+        $order->total_paid_tax_excl = 1500;
+        $order->id_currency = 1;
+        $order->method('getCartProducts')->willReturn($products);
+        $order->method('getProducts')->willReturn($products);
+
+        $orderPresenter = new OrderPresenter();
```
tomjas1997, line 63: could use getService here instead of new

tests/Integration/Subscription/Presenter/OrderPresenterTest.php
```diff
@@ -0,0 +1,74 @@
...
+                'product_price_wt' => 1210.00,
+                'total_wt' => 1210.00,
+                'product_name' => 'test-product-3',
+                'product_quantity' => 3,
+                'product_id' => 3,
+                'id_customization' => 1,
+            ],
+        ];
+
+        $order = $this->createMock(\Order::class);
```
tomjas1997, line 57: mocks and integration testing :(

tomjas1997 (approved):
Looks OK

tests/Integration/Subscription/Presenter/OrderPresenterTest.php
mandan2: Though about just using new \Order(1) but thought that it's easier to prepare data with just creating mock.

PR #817, opened 2023-09-25
PIPRES-261: Get carrier price to create subscription

mandan2:
Subscription order should consist of only subscription product price and selected carrier shipping price. We could not duplicate cart here as it would become abandoned so found another way to try to retrieve carrier price and check if carrier would be available for recurring order.
We used default cart core methods so it retrieved order total with our given product. Gift wrapping would be included too, which could give us false information about carrier zone restrictions so decided write rule to not use gift wrapping as it should not appear in recurring order too.


subscription/Exception/CouldNotProvideSubscriptionCarrierDeliveryPrice.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Mollie\Subscription\Exception;
+
+class CouldNotProvideSubscriptionCarrierDeliveryPrice extends MollieSubscriptionException
+{
+    public static function failedToFindSelectedCarrier(): CouldNotProvideSubscriptionCarrierDeliveryPrice
```
tomjas1997, line 7: self instead of full name

views/templates/admin/Subscription/subscriptions-faq.html.twig
```diff
@@ -92,6 +92,21 @@
     </div>
   </div>
 
+  <div class="row">
+    <div class="col-lg-12">
+      <div class="card">
+        <h3 class="card-header">
+          <i class="material-icons">info_outline</i> {{ giftWrappingTitle }}
```
tomjas1997, line 99: escape

PR #818, opened 2023-09-26
PIPRES-261: Set carrier for recurring order

mandan2:
Trying to apply selected carrier from BO merchant options. On application failure recurring order is not created. On price indifference log is printed out as order was successfully created.


subscription/Handler/RecurringOrderHandler.php
```diff
@@ -182,24 +184,66 @@ private function createSubscription(Payment $transaction, MolRecurringOrder $rec
...
-            ['transaction_id' => $transaction->id],
-            null,
-            false,
-            $newCart->secure_key
+        if (!$carrier) {
+            throw CouldNotHandleRecurringOrder::failedToFindSelectedCarrier();
+        }
+
+        $newCart->setDeliveryOption(
+            [(int) $newCart->id_address_delivery => sprintf('%d,', (int) $carrier->id)]
```
tomjas1997, line 201: is the sprintf really needed here?

subscription/Handler/RecurringOrderHandler.php
```diff
@@ -182,24 +184,66 @@ private function createSubscription(Payment $transaction, MolRecurringOrder $rec
...
         );
 
+        $newCart->update();
+
+        if (sprintf('%d,', (int) $carrier->id) !==
```
tomjas1997, line 206: could just cast into int so comma would not be needed anymore

subscription/Handler/RecurringOrderHandler.php
mandan2: Yes, verified that without sending this exact format changing delivery option won't work.

PR #827, opened 2023-10-16
PIPRES-329: Use Mollie mapped status for initial subscription order

mandan2:
Previously this approach was used but temporary we changed initial status to awaiting to let incorrect pricing orders to be submitted. As awaiting status is not logable, transaction ID is not being inserted into payment, so had to change it back.


subscription/Handler/RecurringOrderHandler.php
```diff
@@ -258,23 +262,30 @@ private function createSubscription(Payment $transaction, MolRecurringOrder $rec
...
+                    'Order price' => $cartTotal,
+                ]);
+            }
+
             $specificPrice->delete();
 
             throw $exception;
         }
 
+        if (!NumberUtility::isEqual($cartTotal, $subscriptionPaidTotal)) {
```
tomjas1997, line 277: Yeah, seems like it should be just prior to validateOrder

tomjas1997 (approved):

PR #840, opened 2023-11-08
PIPRES-319: Lock webhook controller

mandan2:
Locking mechanism for webhook controller. 

Additionally improved a bit webhook error handle logic. Later on all processing should be moved to separate service out of controller.

Release is not called as destruct method will be called to release key. Tried to interrupt program to check what happens without release execution, but everything works fine in the next execution. 

image

### DIFF
--- a/controllers/front/webhook.php
+++ b/controllers/front/webhook.php
@@ -10,11 +10,12 @@
  * @codingStandardsIgnoreStart
  */
 
-use Mollie\Api\Exceptions\ApiException;
 use Mollie\Config\Config;
 use Mollie\Controller\AbstractMollieController;
 use Mollie\Errors\Http\HttpStatusCode;
 use Mollie\Handler\ErrorHandler\ErrorHandler;
+use Mollie\Infrastructure\Adapter\Lock;
+use Mollie\Logger\PrestaLoggerInterface;
 use Mollie\Service\TransactionService;
 use Mollie\Utility\TransactionUtility;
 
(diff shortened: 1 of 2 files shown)


controllers/front/webhook.php
```diff
@@ -38,79 +39,121 @@ class MollieWebhookModuleFrontController extends AbstractMollieController
...
+        }
+
+        /** @var Lock $lock */
+        $lock = $this->module->getService(Lock::class);
+
         try {
-            exit($this->executeWebhook());
+            $lock->create($transactionId);
+
+            $acquired = $lock->acquire();
```
tomjas1997, line 78: $isAcquired would be better name

controllers/front/webhook.php
```diff
@@ -38,79 +39,121 @@ class MollieWebhookModuleFrontController extends AbstractMollieController
...
+            );
+
+            $errorHandler->handle($exception, $exception->getCode(), false);
+
+            $this->respond('failed', HttpStatusCode::HTTP_BAD_REQUEST, 'Failed to process webhook');
+
+            exit;
+        }
+
+        $this->respond('success', HttpStatusCode::HTTP_OK, $result);
```
tomjas1997, line 121: Should release lock after this

controllers/front/webhook.php
```diff
@@ -38,79 +39,121 @@ class MollieWebhookModuleFrontController extends AbstractMollieController
...
+                [
+                    'Exception message' => $exception->getMessage(),
+                    'Exception code' => $exception->getCode(),
+                    'transaction_id' => $transactionId,
+                ]
+            );
+
+            $errorHandler->handle($exception, $exception->getCode(), false);
+
+            $this->respond('failed', HttpStatusCode::HTTP_BAD_REQUEST, 'Failed to process webhook');
```
tomjas1997, line 116: same here

mandan2:
Using file lock mechanism, currently not implementing this one

PR #849, opened 2023-11-27
PIPRES-349: Append subscription carrier ID to metadata

mandan2:
Added currently active subscription carrier ID to Mollie payment's metadata. Added checks to see if currently active carrier is the same as set in the order. Current idea is to have same carrier for all of the subscriptions.

Improved some repositories to use abstractRepository and be injected via interface.


subscription/DTO/SubscriptionCarrierDeliveryPriceProviderData.php
```diff
@@ -0,0 +1,85 @@
...
+ * @codingStandardsIgnoreStart
+ */
+
+namespace Mollie\Subscription\DTO;
+
+if (!defined('_PS_VERSION_')) {
+    exit;
+}
+
+class SubscriptionCarrierDeliveryPriceProviderData
```
tomjas1997, line 19: "Provider" does  not seem needed

subscription/Exception/CouldNotCreateSubscriptionData.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Mollie\Subscription\Exception;
+
+if (!defined('_PS_VERSION_')) {
+    exit;
+}
+
+class CouldNotCreateSubscriptionData extends MollieSubscriptionException
+{
+    public static function failedToFindMolCustomer(string $email): self
```
tomjas1997, line 21: Mollie

PR #857, opened 2023-12-05
PIPRES-349: Abstract subscription order total provider

mandan2:
This provider will be used in other service too so abstracted it to reduce code repetition.

### DIFF
--- a/subscription/Provider/SubscriptionOrderAmountProvider.php
+++ b/subscription/Provider/SubscriptionOrderAmountProvider.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * Mollie       https://www.mollie.nl
+ *
+ * @author      Mollie B.V. <info@mollie.nl>
+ * @copyright   Mollie B.V.
+ * @license     https://github.com/mollie/PrestaShop/blob/master/LICENSE.md
+ *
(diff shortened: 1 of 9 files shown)


subscription/Provider/SubscriptionOrderAmountProvider.php
```diff
@@ -0,0 +1,73 @@
...
+                    $data->getCustomerId(),
+                    $data->getSubscriptionProduct(),
+                    $data->getSubscriptionCarrierId()
+                )
+            );
+        } catch (\Throwable $exception) {
+            throw CouldNotProvideSubscriptionOrderAmount::failedToProvideCarrierDeliveryPrice($exception);
+        }
+
+        $orderTotal = (float) $data->getSubscriptionProduct()['total_price_tax_incl'] + $deliveryPrice;
```
tomjas1997, line 60: Not sure if you have numberUtility in this project, but maybe adding up floats would be best done with that library.

tomjas1997, line 60: no posibility of 'total_price_tax_incl' as a getter instead of an array?

tomjas1997 (approved):

PR #858, opened 2023-12-05
PIPRES-349: Subscription carrier data provider

mandan2:
Prepared data to update to be transferred via API


subscription/Provider/UpdateSubscriptionCarrierDataProvider.php
```diff
@@ -0,0 +1,105 @@
...
+                )
+            );
+        } catch (\Throwable $exception) {
+            throw CouldNotProvideSubscriptionCarrierData::failedToProvideSubscriptionOrderAmount($exception);
+        }
+
+        return new UpdateSubscriptionData(
+            $data->getMollieCustomerId(),
+            $data->getMollieSubscriptionId(),
+            null,
```
tomjas1997, line 100: Not sure about this null if it's really needed in any way

tomjas1997 (approved):

PR #861, opened 2023-12-11
PIPRES-349: Removed redundant services to optimize process

mandan2:
By writing the rest of the services noticed that we will need a lot of the data, which is not passable currently. By investigating noticed that we could reduce separate service calling to make data less coupled.


tests/Integration/Factory/MolRecurringOrderFactory.php
```diff
@@ -0,0 +1,43 @@
...
+ * @codingStandardsIgnoreStart
+ */
+
+namespace Mollie\Tests\Integration\Factory;
+
+class MolRecurringOrderFactory implements FactoryInterface
+{
+    public static function create(array $data = []): \MolRecurringOrder
+    {
+        $recurringOrder = new \MolRecurringOrder();
```
tomjas1997, line 19: Could start using our Factory package for these :D

tomjas1997 (approved):

PR #865, opened 2023-12-18
PIPRES-349: Use same shared abstract repository

mandan2:
Moved abstract repository with interface to shared namespace to reduce repetitive code.

Later implement findOrFail into all of carrier change repository calls.


controllers/front/subscriptions.php
```diff
@@ -52,40 +52,33 @@ class mollieSubscriptionsModuleFrontController extends ModuleFrontController
...
         /** @var RecurringOrdersPresenter $recurringOrdersPresenter */
         $recurringOrdersPresenter = $this->module->getService(RecurringOrdersPresenter::class);
 
-        $molCustomer = $molCustomerRepository->findOneBy(['email' => $context->customer->email]);
+        try {
+            /** @var \MolCustomer $molCustomer */
+            $molCustomer = $molCustomerRepository->findOrFail([
+                'email' => $this->context->customer->email,
+            ]);
+        } catch (\Throwable $exception) {
```
tomjas1997, line 75: Does not seem needed, maybe findOneBy was good enough?

shared/Infrastructure/Repository/AbstractRepository.php
```diff
@@ -0,0 +1,106 @@
...
+class AbstractRepository implements ReadOnlyRepositoryInterface
+{
+    /**
+     * @var string
+     */
+    private $fullyClassifiedClassName;
+
+    public function __construct($fullyClassifiedClassName)
+    {
+        if (is_object($fullyClassifiedClassName)) {
```
tomjas1997, line 30: Is it ever an object? Pretty sure we are passing Model::class

tomjas1997: Suggestion - 
if (is_object($fullyClassifiedClassName)) { $fullyClassifiedClassName = get_class($fullyClassifiedClassName)} , early return does not really do much here

shared/Infrastructure/Repository/AbstractRepository.php
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        $all = $psCollection->getAll();
+
+        /* @phpstan-ignore-next-line */
+        return false === $all ? null : $all;
+    }
+
+    /** {@inheritdoc} */
+    public function findAllByOrFail(array $keyValueCriteria, int $langId = null): \PrestaShopCollection
```
tomjas1997, line 76: Does not seem to fit the logic, for singular - it's OK, but for multiple there might be 1 out of 3 found and would still pass this check, so kind of unnecessary

shared/Infrastructure/Repository/AbstractRepository.php
```diff
@@ -0,0 +1,106 @@
...
+        return $records;
+    }
+
+    /** {@inheritdoc} */
+    public function findOrFail(array $keyValueCriteria, int $langId = null): \ObjectModel
+    {
+        try {
+            $value = $this->findOneBy($keyValueCriteria, $langId);
+        } catch (\Throwable $exception) {
+            throw CouldNotHandleAbstractRepository::unknownError($exception);
```
tomjas1997, line 97: Could create a better named exception, like MollieDatabaseException and MollieModelNotFoundException

src/Service/MailService.php
```diff
@@ -132,8 +132,25 @@ public function sendOrderConfMail(Order $order, $orderStateId)
...
     public function sendSubscriptionCancelWarningEmail(int $recurringOrderId): void
     {
-        $recurringOrder = $this->recurringOrderRepository->findOneBy(['id_mol_recurring_order' => $recurringOrderId]);
-        $recurringOrderProduct = $this->recurringOrdersProductRepository->findOneBy(['id_mol_recurring_orders_product' => $recurringOrder->id_mol_recurring_orders_product]);
+        try {
+            /** @var \MolRecurringOrder $recurringOrder */
+            $recurringOrder = $this->recurringOrderRepository->findOrFail([
+                'id_mol_recurring_order' => $recurringOrderId,
+            ]);
+        } catch (\Throwable $exception) {
```
tomjas1997, line 140: If you don't need to perform some action with this exception - there's no need to catch it, leave it to upper level to handle

src/Service/MailService.php
```diff
@@ -132,8 +132,25 @@ public function sendOrderConfMail(Order $order, $orderStateId)
...
+            // TODO refactor this with data provider
+            return;
+        }
+
+        try {
+            /** @var \MolRecurringOrdersProduct $recurringOrderProduct */
+            $recurringOrderProduct = $this->recurringOrdersProductRepository->findOrFail([
+                'id_mol_recurring_orders_product' => $recurringOrder->id_mol_recurring_orders_product,
+            ]);
+        } catch (\Throwable $exception) {
```
tomjas1997, line 150: Same here

subscription/Handler/RecurringOrderHandler.php
```diff
@@ -110,10 +111,22 @@ public function __construct(
...
         $transaction = $this->mollie->getApiClient()->payments->get($transactionId);
-        $recurringOrder = $this->recurringOrderRepository->findOneBy(['mollie_subscription_id' => $transaction->subscriptionId]);
+
+        try {
+            /** @var \MolRecurringOrder $recurringOrder */
+            $recurringOrder = $this->recurringOrderRepository->findOrFail([
+                'mollie_subscription_id' => $transaction->subscriptionId,
+            ]);
+        } catch (\Throwable $exception) {
+            throw TransactionException::unknownError($exception);
```
tomjas1997, line 127: Not sure about this exception, it is really necessary? All of the information is already inside the initial exception, so it can just be thrown to upper level.

controllers/front/subscriptions.php
mandan2: No, this is quite useful. No need to check separately for not found customer, so avoided one if condition.

shared/Infrastructure/Repository/AbstractRepository.php
mandan2: renamed to MollieDatabaseException

controllers/front/subscriptions.php
tomjas1997: $this->prepareTemplate(
$customer ? $recurringOrdersPresenter->present($molCustomer->customer_id) : []
);

saves 4 in your case :D

tomjas1997 (approved):

PR #905, opened 2024-04-09
PRES-430: Make cloudsync option to install

JevgenijVisockij:


tomjas1997 (approved):
Not really sure about this change, but i guess its needed for some reason

PR #923, opened 2024-05-14
PIPRES-349

JevgenijVisockij:


JevgenijVisockij:
@todo change upgrade version to the one from release

PR #931, opened 2024-05-27
PIPRES-431/ League Container fix

GytisZum:
League container fix


sladkova:
@GytisZum is seems this fix make impossible to pass empty (``) phone number to the OrderData which in turn breaks usage of the Orders API - we send `N/A` for empty phone numbers and getting `The phone field should contain  a valid phone number, in E 164 format` response

PR #946, opened 2024-06-26
PIPRES-445/http status

GytisZum:


Makefile
```diff
@@ -90,9 +90,4 @@ npm-package-install:
 prepare-zip:
 	composer install --no-dev --optimize-autoloader --classmap-authoritative
 	composer dump-autoload --no-dev --optimize --classmap-authoritative
-	cp .github/.htaccess vendor/.htaccess
```
aleksiuno, line 93: Please doublecheck if .htaccess file is not needed as it could introduce security issues in module vendor folder

aleksiuno (approved):

Makefile
GytisZum: .htaccess is not excluded and always exist, this copy was extra step which was doing nothing at module preparation script

Issue #965, opened 2024-08-28
Subscription is not created if product (combination) has other attribute(s) than Mollie subscription one


zorrobiwan:
### Expected behavior
One product with combinations (for example for these attributes : "Colors" and "Mollie subscription").
When creating an order for a combination with these attribute values "Red" and "Daily", the subscription is not created and error is logged if debug is enabled
`Log ID (66cf166be3aa86.38147825) | Failed to create recurring order . context: {\"Exception message\":\"Failed to create subscription data.\",\"Exception code\":1014}`
Order is well created.

Order AND subscription are created if the combination has only the "Mollie Subscription" attribute.

### Actual behavior
Order AND subscription should be created even combination has other attributes than "Mollie Subscription"

### Steps to reproduce the behavior
Create a product and combination with multiple attributes (ex: "Colors" and "Mollie subscription").
Create an order and check your subscription which is not created.
If Mollie debug is enabled, check the logs and see the error message.

### Environment
* PrestaShop version: 1.7.7.6
* Mollie module version: 6.2.2
* PHP version: 7.3.31-1~deb10u7

Extra debug log : 

```
Mollie\\Subscription\\Exception\\SubscriptionIntervalException: No interval exists for this 241 attribute in /var/www/html/staging/modules/mollie/subscription/Provider/SubscriptionStartDateProvider.php:72
Stack trace:
#0 /var/www/html/staging/modules/mollie/subscription/Factory/CreateSubscriptionDataFactory.php(159): Mollie\\Subscription\\Provider\\SubscriptionStartDateProvider->getSubscriptionStartDate(Object(Combination))
#1 /var/www/html/staging/modules/mollie/subscription/Handler/SubscriptionCreationHandler.php(85): Mollie\\Subscription\\Factory\\CreateSubscriptionDataFactory->build(Object(Order), Array)
#2 /var/www/html/staging/modules/mollie/src/Handler/Order/OrderCreationHandler.php(294): Mollie\\Subscription\\Handler\\SubscriptionCreationHandler->handle(Object(Order), \'creditcard\')
#3 /var/www/html/staging/modules/mollie/src/Handler/Order/OrderCreationHandler.php(164): Mollie\\Handler\\Order\\OrderCreationHandler->createRecurringOrderEntity(Object(Order), \'creditcard\')
[...]

PR #966, opened 2024-08-28
Go through all combination attributes to check subscription interval

zorrobiwan:
Solve https://github.com/mollie/PrestaShop/issues/965


GytisZum (changes requested):
LGTM! :1st_place_medal:  Small fixes needed for PHP code standard (linter/cs-fixer fixes) and define `$interval` variable above switch method.

PR #1026, opened 2025-01-14
BUGFIX async webhook fix

MarijusCoding:
| Questions       | Answers
| Branch?         | release-6.2.6
| Description?    | Fixed issue with latest mollie API changes that made webhook async.
| Type?           | bug fix
| How to test?    | Try checkouting with different statuses
| Fixed issue ?   | PIPRES-526

### DIFF
--- a/controllers/front/return.php
+++ b/controllers/front/return.php
@@ -204,6 +204,7 @@ class MollieReturnModuleFrontController extends AbstractMollieController
     protected function processGetStatus()
     {
         header('Content-Type: application/json;charset=UTF-8');
+
         /** @var PaymentMethodRepository $paymentMethodRepo */
         $paymentMethodRepo = $this->module->getService(PaymentMethodRepository::class);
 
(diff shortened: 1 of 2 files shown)


controllers/front/return.php
```diff
@@ -103,7 +103,8 @@ public function initContent()
             // any paid payments for this cart?
 
             if (false === $data['mollie_info']) {
-                $data['mollie_info'] = $paymentMethodRepo->getPaymentBy('order_id', (int) Order::getOrderByCartId($idCart));
+                $orderId = (int) Order::getOrderByCartId($idCart);
```
tomas862, line 106: does mollie support many PS versions ? If I not mistaken this static call in one version exists, in another, named differently. Can someone confirm or deny this ?

controllers/front/return.php
```diff
@@ -103,7 +103,8 @@ public function initContent()
...
+                $data['mollie_info'] = $orderId != 0 ? $paymentMethodRepo->getPaymentBy('order_id', (int) $orderId) : [];
             }
             if (false === $data['mollie_info']) {
```
tomas862, line 109: () $orderId, you already casted it once.

tomas862 (changes requested):

controllers/front/return.php
MarijusCoding: Yeah, this method is deleted in PS 9 and we already have PR which replaces all those deprecated methods. Thank you, I replaced it with new one :pray:

controllers/front/return.php
MarijusCoding: Thank you :pray:

GytisZum (approved):

PR #1091, opened 2025-03-24
PIPRES-559 trustly rename in BO

MarijusDilys:
| Questions       | Answers
| Branch?         | Develop
| Description?    | Trustly called "Pay by Bank", so we renamed it to be "Pay by Bank (Trustly)"
| Type?           | bug fix / improvement / new feature / refactoring
| How to test?    | Indicate how to verify that this change works as expected.
| Fixed issue ?   | If none leave blank


MarijusDilys:
image
This is only used to admin configurations page so its right spot @GytisZum

GytisZum (approved):

PR #1105, opened 2025-04-22
PIPRES-576 Remove unsupported characters in Product URL

MarijusDilys:
| Questions       | Answers
| Branch?         | Master/Release
| Description?    | Please be specific when describing the PR. <br> Every detail helps: versions, browser/server configuration, specific module/theme, etc. Feel free to add more information below this table.
| Type?           | bug fix / improvement / new feature / refactoring
| How to test?    | Indicate how to verify that this change works as expected.
| Fixed issue ?   | If none leave blank

### DIFF
--- a/src/Utility/TextFormatUtility.php
+++ b/src/Utility/TextFormatUtility.php
@@ -11,6 +11,8 @@
 
 namespace Mollie\Utility;
 
+use Transliterator;
+
 if (!defined('_PS_VERSION_')) {
     exit;
 }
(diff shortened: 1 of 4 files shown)


src/Utility/TextFormatUtility.php
```diff
@@ -21,4 +23,18 @@ public static function formatNumber($unitPrice, $apiRoundingPrecision, $docPoint
...
+    /**
+     * Replace all accented chars by their equivalent non-accented chars.
+     *
+     * @param string $string
+     *
+     * @return string
+     */
+    public static function replaceAccentedChars(string $string): string
+    {
+        $transliterator = Transliterator::create('Any-Latin; Latin-ASCII');
```
GytisZum, line 36: Maybe just in case check is this class exist and then proceed otherwise just return same string
```suggestion
        if (class_exists('Transliterator', false) && $transliterator = \Transliterator::create($string)) {
```

GytisZum (approved):

Issue #1410, opened 2026-06-11
PHP 8.3 deprecations in module code (number_format(null), jsonSerialize) — fatal TypeError on PHP 9 — v6.4.3


rcoenen:
### Summary

On PrestaShop 9.1 / PHP 8.3 with Mollie module **v6.4.3**, two PHP deprecation notices are emitted from the module's **own** source (not a vendored dependency) when the payment controller runs. They are harmless on PHP 8.x with `display_errors` off, but `number_format(null)` becomes a **fatal `TypeError` on PHP 9**, so this is a forward-compatibility blocker rather than just noise.

Related but distinct (vendored dependency, already closed): #640.

### Environment

| | |
|---|---|
| Mollie module | **6.4.3** (latest release) |
| PrestaShop | **9.1.0** |
| PHP | **8.3.31** |

### Steps to reproduce

1. Run the module on PHP 8.3 with `display_errors = On` (e.g. PrestaShop dev mode).
2. Start a payment, e.g. `GET /module/mollie/payment?method=ideal&rand=...`.

### Actual behavior

```
Deprecated: number_format(): Passing null to parameter #1 ($num) of type float is deprecated
in modules/mollie/src/Utility/TextFormatUtility.php on line 24

Deprecated: Return type of Mollie\DTO\Object\Payment::jsonSerialize() should either be compatible
with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be
used to temporarily suppress the notice in modules/mollie/src/DTO/Object/Payment.php on line 111
```

### Root cause + suggested fix

**1. `src/Utility/TextFormatUtility.php:24`** — `$unitPrice` can be `null`:

```php
public static function formatNumber($unitPrice, $apiRoundingPrecision, $docPoint = '.', $thousandSep = '')
{
    return number_format($unitPrice, $apiRoundingPrecision, $docPoint, $thousandSep);
}
```

Passing `null` to `number_format()` is deprecated in 8.1+ and a `TypeError` in PHP 9. Minimal fix:

```php
return number_format((float) $unitPrice, $apiRoundingPrecision, $docPoint, $thousandSep);
```

(It may also be worth tracing *why* a null amount reaches `formatNumber()` in the payment path — that could be a separate upstream issue — but the cast removes the forward-compat break.)

**2. `src/DTO/Object/Payment.php:111`** — `jsonSerialize()` has no compatible return type:

```php
public function jsonSerialize()
```

[...]

PR #1411, opened 2026-06-11
Fix PHP 8.1+ deprecations on the payment path (number_format, jsonSerialize)

rcoenen:
Fixes the PHP 8.1+ deprecation notices reported in #1410. They surface on the payment/checkout path on PHP 8.3 and become a **fatal `TypeError` on PHP 9**, so this is a forward-compatibility fix (no behaviour change on supported PHP versions).

### Changes

**1. `number_format(null)` — `src/Utility/TextFormatUtility.php`**

`formatNumber()` can receive a `null` `$unitPrice`, and `number_format(null, …)` is deprecated since PHP 8.1 and a `TypeError` on PHP 9. Casting to `float` (null → `0.0`) preserves the current output — `number_format(null, …)` already yields `"0"` — with no behaviour change.

**2. `jsonSerialize()` return type — 4 DTOs**

`JsonSerializable::jsonSerialize()` is declared `: mixed` since PHP 8.1, so implementations with no declared return type emit a deprecation. Added `#[\ReturnTypeWillChange]` (parsed as a comment on PHP 7.x, recognised on 8.1+) to the four implementations that declare no return type:

- `src/DTO/Object/Payment.php` (the one observed firing in #1410)
- `src/DTO/Object/Company.php`
- `src/Errors/Error.php`
- `src/DTO/ApplePay/Carrier/Carrier.php`

The four DTOs that already declare `: array` (`OrderLine`, `OrderData`, `PaymentData`, `PaymentLine`) are covariant with `mixed` and were left untouched.

### Notes

- `#[\ReturnTypeWillChange]` was chosen over a `: mixed` return type because `composer.json` declares no `php` floor (the module still targets PHP 7.x), and `: mixed` requires PHP 8.0+.
- It may be worth separately investigating *why* a `null` amount reaches `formatNumber()` in the payment flow, but the cast removes the deprecation/`TypeError` regardless.
- `php -l` is clean on all changed files.

Refs #1410

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2073,9 +2073,9 @@
       "dev": true
     },
     "node_modules/nanoid": {
-      "version": "3.3.7",
-      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.7.tgz",
-      "integrity": "sha512-eSRppjcPIatRIMC1U6UngP8XFcz8MQWGQdt1MTBQ7NaAmvXDfvNxbvWV3x2y6CdEUciCSsDHDQZbhYaB8QEo2g==",
+      "version": "3.3.11",
+      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.11.tgz",
+      "integrity": "sha512-N8SpfPUnUp1bK+PMYW8qSWdl9U+wwNWI4QKxOYDy9JAro3WMX7p2OeVRF9v+347pnakNevPmiHhNmZ2HbFA76w==",
       "dev": true,
       "funding": [
         {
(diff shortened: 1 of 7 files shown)


TLabutis:
Thanks for this, and sorry it sat so long. Closing it, but the fix itself was right and it is being shipped.

Two reasons this branch cannot be merged as is:

1. It targets `release-6.4.4`, which is no longer maintained (last commit 2026-07-01). Fixes now go to `develop`.
2. Alongside the 5 source files it carries ~7400 lines of unrelated lockfile churn in `package-lock.json` and `views/assets/package-lock.json`.

The work now lives on #1509 (PIPRES-814), against `develop`, touching exactly the same 5 files you identified plus a unit test for `TextFormatUtility`. So your diagnosis is landing, just on the current base and without the lockfile noise.

Appreciated the report.

Issue #1428, opened 2026-06-29
Order actions (refund/capture/ship/cancel) do nothing on PS 1.7.7+ Symfony order page


PrestaEdit:
### Expected behavior

On the BO order detail page in PS 1.7.7+ (Symfony route `/sell/orders/{id}/view`), clicking the Mollie order-management buttons (**Initiate Refund**, Capture, Ship, Cancel) should open the corresponding confirmation modal and perform the action.

### Actual behavior

The Mollie panel and its buttons render correctly (via `hookDisplayAdminOrder`), but clicking them does **nothing** — no modal, no error message.

`views/js/admin/order_info.js` and its `Media::addJsDef` config (`transaction_id`, `resource`, `ajax_url`, `order_id`, `trans`) are never loaded on the page, so the click handler throws a silent `ReferenceError: transaction_id is not defined`.

**Root cause:** in `mollie.php::hookActionAdminControllerSetMedia`, the order-detail assets are loaded only inside a block guarded by:

```php
if ('AdminOrders' === Tools::getValue('controller') && Tools::getValue('id_order')) { ... }
```

On the PS 1.7.7+ Symfony order route, neither `controller` nor `id_order` is a GET parameter, so this block never runs and the assets are never enqueued. (By contrast, `hookDisplayBackOfficeHeader` correctly uses `$this->context->controller->controller_name`.)

Confirmed on `master` and `v6.4.4-beta`. Note: `src/Loader/OrderManagementAssetLoader` looks intended to load these assets but is never called anywhere.

### Steps to reproduce the behavior

1. PrestaShop 1.7.7+ (new Symfony order page).
2. Open a paid Mollie order in the BO: `/sell/orders/{id}/view`.
3. Click **Initiate Refund** → nothing happens.
4. In the browser console: `typeof transaction_id` → `"undefined"`, and `order_info.js` is absent from `document.scripts`.

### Environment

* PrestaShop version: 1.7.7.0
* Mollie module version: 6.4.3 (also reproduced on `master` and `v6.4.4-beta`)
* PHP version: 7.4

### Additional note

`views/css/admin/order_info.css` is `addCSS`'d in the same code path but does not exist in the package (404 on the order page).

TLabutis:
Hey, @PrestaEdit

Thanks for the detailed report and the PR, much appreciated.

[...]

PR #1429, opened 2026-06-29
Fix: order management actions do nothing on PS 1.7.7+ Symfony order page

PrestaEdit:
| Questions       | Answers
| Branch?         | Master
| Description?    | Fixes the Mollie order-management actions (Refund / Capture / Ship / Cancel) doing nothing on the PS 1.7.7+ Symfony order page. See details below.
| Type?           | bug fix
| How to test?    | On PS 1.7.7+, open a paid Mollie order in the BO (`/sell/orders/{id}/view`). Before: clicking **Initiate Refund** does nothing and `typeof transaction_id` is `"undefined"` in the console. After: the confirmation modal opens and `transaction_id` is defined.
| Fixed issue ?   | #1428

### Problem

`views/js/admin/order_info.js` and its `Media::addJsDef` config (`transaction_id`, `resource`, `ajax_url`, `order_id`, `trans`) are enqueued only inside `mollie.php::hookActionAdminControllerSetMedia`, behind:

```php
if ('AdminOrders' === Tools::getValue('controller') && Tools::getValue('id_order')) { ... }
```

On the PS 1.7.7+ Symfony order route (`/sell/orders/{id}/view`), neither `controller` nor `id_order` is a GET parameter, so this block never runs. The script is never loaded, the globals are never defined, and the click handler throws a silent `ReferenceError`. The buttons (rendered by `hookDisplayAdminOrder`) therefore do nothing. Reproduced on `master` and `v6.4.4-beta`.

### Fix

Load `order_info.js` and its config from `hookDisplayAdminOrder`, which already receives `$params['id_order']` and runs on the Symfony order page. The config is emitted as safe JSON (`JSON_HEX_*`) into the template, which defines the globals and then loads the script. An idempotency guard in `order_info.js` prevents double-binding when the legacy media path also loads it.

This keeps the existing JS contract (bare globals) untouched and is the minimal change to make the actions work on Symfony-routed order pages.

### DIFF
--- a/mollie.php
+++ b/mollie.php
@@ -702,6 +702,35 @@ class Mollie extends PaymentModule
                 'isCaptured' => $isCaptured,
                 'isShipped' => $isShipped,
                 'isCanceled' => $isCanceled,
+                // Load order_info.js and its config from this hook, which receives the order id and
+                // runs on the PS 1.7.7+ Symfony order page (/sell/orders/{id}/view).
+                // hookActionAdminControllerSetMedia is guarded by Tools::getValue('controller') and
+                // Tools::getValue('id_order'), both empty on a Symfony route, so the order management
+                // JS was never enqueued there and the action buttons did nothing.
+                'mollie_order_info_js' => $this->getPathUri() . 'views/js/admin/order_info.js',
+                'mollie_order_info_config' => json_encode([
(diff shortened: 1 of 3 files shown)


TLabutis:
PR is closed as we were not able to replicate the issue in our testing environments.

Issue #1434, opened 2026-07-04
Total_paid_real stays 0.0 after Bank payment comes through


twazerty:
Yesterday we ran into an issue when somebody paid via Bank instead of iDeal. We retrieve orders through the prestashop orders api. The field total_paid_real should contain the actual amount that is paid. This works properly for iDeal payments. The order is created after payment is made. However when a user wants to pay via Bank, the order is created with the status Awaiting Mollie Payment. At this point the total_paid_real is still at 0.0. However when the user pays to Mollie and you update the order, the total_paid_real stays at 0.0.

We recently updated the plugin from 6.2.8.1 to 6.4.3. We are not 100% sure, but we think this worked properly in the previous version. We don't get many payments via Bank through Mollie, so we can be wrong.

If I use the test mode of the plugin I have the same results. total_paid_real stays at 0.0

Is this a bug in the plugin? I could only find #309, but reversed issue back then.

TLabutis:
Hi,

Thank you for taking the time to report this.

Our next step is to investigate and replicate the issue mentioned. We are going to inform you as soon as i have more details. 

Thanks again.

--
Best Regards,
Invertus Support team.

TLabutis:
Hi, @twazerty

to confirm the cause of the Bank transfer total_paid_real issue, could you check two things for us?

1. Which exact PrestaShop version are you running? You can find it in the Back Office footer, or under Advanced Parameters > Information.
2. Is invoice generation enabled or disabled on your shop? Please check Back Office > Shop Parameters > Order Settings > Invoices > "Enable invoices". Let us know if it is set to Yes or No.

--
Best Regards,
Invertus Support team.

twazerty:
Hi @TLabutis,

1. Prestashop 8.2.0
2. The setting is not there, but at (Main categorie SELL) -> Orders -> Invoices -> Enable Invoices. It is set to true (but maybe we want to disable this by the way).

TLabutis:
Hi,

We are passing this additional information to our development team to investigate further.

Thanks for your cooperation.


--
Best Regards,
Invertus Support team.

TLabutis:
Hi, @twazerty 

We have prepared a patched version of the module (6.4.4) that includes the fix for the total_paid_real issue on bank transfer orders.

Could you please test it on a staging or test copy of your shop (not production) to confirm it resolves the issue on your setup:

1. Back up your current module / test on a non-production shop.
[...]

PR #1438, opened 2026-07-09
PIPRES-793 Fix total_paid_real not updated on bank transfer orders

TLabutis:
## Problem
For bank transfer orders, `total_paid_real` stays `0.0` even after the payment is completed and the order is marked as paid. iDEAL is unaffected. This surfaces when reading orders via the PrestaShop Orders API (webservice). Reported in #1434 (PIPRES-793).

## Root cause
Bank transfer orders are created in a non-logable "awaiting" state, so PrestaShop core does not record an order payment at creation and `total_paid_real` starts at 0. When the payment is later completed on the webhook, `TransactionService::updateTransaction()` records the payment but never maintains `total_paid_real`, so the paid order keeps `total_paid_real = 0`. This happens regardless of the shop "Enable invoices" setting.

## Fix
Recompute `total_paid_real` from the order's own order payments (currency-matched) at the end of `updateTransaction()`, so it is always maintained no matter how the payment was recorded. The recompute is guarded to be idempotent and does not change orders that are already correct (e.g. iDEAL, created directly in a paid state).

## Testing
Verified end-to-end on PrestaShop 8.2.1 + Mollie test mode with "Enable invoices" ON:

| Method | Before fix | After fix |
|---|---|---|
| Bank transfer | total_paid_real = 0.00 | total_paid_real = 22.94 ✅ |
| iDEAL (regression) | 22.94 | 22.94 ✅ (unchanged) |

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 # Changelog #
 
 ## Changes in release 6.4.3
++ Fixed total_paid_real staying 0 on bank transfer orders after the payment is completed
 + Fixed Payment API race condition causing false "payment failed" errors
 + Fixed rounding distribution for whole-number differences in order line amounts
 + Fixed Apple Pay Direct payment when country is not active in shop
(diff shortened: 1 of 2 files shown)


TLabutis:
Superseding with a PR based on master (6.4.4, the current released line). develop is behind at 6.4.3, so the fix should target the released code the merchant runs.

PR #1446, opened 2026-07-20
PIPRES-779: fix "On backorder (paid)" status for orders placed at zero stock

sarasko:
| Questions       | Answers
| Branch?         | Master
| Description?    | Fixes backorder detection so an order placed at exactly 0 stock is set to "On backorder (paid)".
| Type?           | bug fix
| How to test?    | See "How to test" below.
| Fixed issue ?   | PIPRES-779

## What was wrong

An order placed when a product sits at exactly **0 stock** (with backorders allowed and `PS_STOCK_MANAGEMENT` on) never moved to **"On backorder (paid)"** — it stayed on "Payment accepted".

Two separate causes:

1. **Wrong comparison.** `OrderStatusService::isOrderBackOrder()` only treated stock *below* zero as a backorder (`product_quantity_in_stock < 0`). A product starting at 0 never goes negative, so it was missed. This is effectively a regression: the same fix lived in PR #383 (2021) and was lost when the logic moved from `TransactionService` to `OrderStatusService`.
2. **Instant-payment methods skipped the check entirely.** For iDEAL, cards, etc., `OrderCreationHandler::createOrder()` set the paid status directly, so the backorder transform never ran at all.

## The fix

- `isOrderBackOrder()` now compares in-stock quantity against the **ordered** quantity, not just against zero. The core decision is extracted into a pure, unit-tested `OrderStatusService::isBackOrder()`.
- `createOrder()` now creates paid orders in the **awaiting** state and moves them to paid through `OrderStatusService`, so the backorder transform runs for instant-payment methods too. This mirrors the bank-transfer and payment-fee flows that already work. The awaiting state has `sendEmail = false`, so no extra customer email is sent.
- **Authorizable payments** (manual capture) keep their integer status — an authorized, uncaptured order must never be turned into a "paid" backorder status.

## How to test

Preconditions: `PS_STOCK_MANAGEMENT` on, a product with **stock 0** and backorders allowed.

1. **Order-before-payment (bank transfer):** place + pay an order for the stock-0 product → order lands on **"On backorder (paid)"**.
2. **Order-after-payment (iDEAL / card):** same → order lands on **"On backorder (paid)"** (previously "Payment accepted").
3. **Regression:** place an order for an **in-stock** product → normal paid status, not backorder.
4. Repeat 1–2 with a payment fee configured → still lands on backorder.

## Tests

- **Unit** `tests/Unit/Service/OrderStatusServiceTest.php` — pins the quantity comparison, including the stock-0 boundary and the in-stock regression guard (7 cases). Runs in CI (`--testsuite Unit`).
- **Integration** `tests/Integration/Service/OrderStatusServiceBackOrderTest.php` — builds a real cart + product + order and runs the actual `OrderStatusService`. Asserts a stock-0 order lands on "On backorder (paid)" (PS state 9) and an in-stock order stays on paid (state 2). Verified locally: **fails on the parent commit** (order lands on state 2) and **passes with the fix**. Not part of the CI Unit job (the integration suite isn't wired into CI here), so it also carries a one-line `CustomerFactory` fix (hash the password) so the integration suite can create customers on PrestaShop 8+.
- `delivery-playbook classify`: TDD required (done); Browser QA / Security / Three-lens all N/A for this diff (no UI, no sensitive surface).
- phpstan was not run locally (not installed in the working checkout); relying on CI. The diff is type-simple (a pure `bool` helper; existing signatures unchanged).

## Needs human verification

- **Contribution / licensing.** The original fix idea came from a merchant-supplied patch (ticket ref "AS2000"). This PR is an independent re-implementation, but per the ticket, contribution consent should be confirmed before merge.
- **Live Mollie round-trip.** The integration test drives order creation + the status transition directly; it does not exercise the full Mollie webhook round-trip (that needs a live shop + reachable webhook). `createOrder`'s change reduces to "validate at awaiting, then `setOrderStatus('paid')`", which the integration test covers at the status level — but a real paid iDEAL/card order on a live shop is worth one confirmation pass before release.
- **Changelog.** Entry added under a new **6.4.5** section (6.4.4 is already released).

Confidence: 85% — the fix is covered by a unit test and by an integration test that builds a real order and reproduces the bug on the parent commit. Remaining gap is the live Mollie webhook round-trip, which this change does not alter.


sarasko:
### Review status

- CI is green across all checks: php-cs-fixer, ESLint, Unit tests, PHPStan (1.7.6.8 / 1.7.7.0 / nightly-apache), and the ZIP build.
- No review comments to address — none were raised by any reviewer or bot.
- **`gemini-code-assist` did not post a review within the 5-minute window.** This is often a repo config issue (the bot isn't installed here, or the reviewer name is misconfigured) rather than an all-clear. Flagging so a maintainer can confirm; proceeding since CI is green and there are no outstanding comments.

### Verification done
- Unit test pins the quantity comparison, incl. the stock-0 boundary and the in-stock regression guard.
- Integration test builds a real order: fails on the parent commit (order lands on "Payment accepted") and passes with the fix (lands on "On backorder (paid)"). Also confirmed manually in the back office.

### Still needs a human before merge
- Contribution/licensing sign-off for the merchant-supplied patch idea.
- A live Mollie webhook round-trip on a real shop (not reproducible locally).

TLabutis (approved):

PR #1452, opened 2026-07-23
PIPRES-792: harden Mollie module against PrestaShop security requirements

justinas389:
| Questions     | Answers |
|---------------|---------|
| Branch?       | Master |
| Description?  | Security review remediation for the Mollie module against the PrestaShop technical development standards. Fixes the confirmed issues that are safe to land now; larger refactors are deferred to follow-up tickets. Full findings (file + line + severity) are in the PDF attached to PIPRES-792. |
| Type?         | improvement (security hardening) |
| How to test?  | See "How to test" below. |
| Fixed issue ? | PIPRES-792 |

## What changed

**SQL injection (1 Medium + 8 Low)**
- `PaymentMethodRepository::getMethodsForCheckout` used `pSQL()` on the numeric `live_environment` column (unquoted → injectable if the value were ever non-int). Switched to `(int)`.
- Added explicit `(int)` casts / `pSQL` to internal values interpolated into SQL in `AttributeRepository`, `CustomerRepository`, `PaymentMethodRepository`, `TaxRulesGroupRepository`, `TaxRuleRepository`, `PrestashopLoggerRepository`, `subscription/Repository/RecurringOrderRepository`, and `mollie.php` (defense-in-depth; these were protected only by PHP type hints).

**AJAX / access control**
- **High (IDOR)** `controllers/front/recurringOrderDetail.php` — `updatePaymentMethod()` and `cancelSubscription()` acted on a POST `recurring_order_id` with only a shop-wide CSRF token (skipped entirely when `PS_TOKEN_ENABLE` is off) and no ownership check. A logged-in customer could cancel or change the payment method of another customer's subscription by guessing a sequential id. Added `customerOwnsRecurringOrder()`, which verifies the record's `id_customer` matches the logged-in customer before either action (mirrors the check already in `initContent()`).
- **Low** `controllers/admin/AdminMollieAjaxController::validateLogo` — the jpg/png guard checked the fixed destination path instead of the uploaded file, so it was inert. Now validates the uploaded file's extension, upload error, and `getimagesize()` dimensions before `move_uploaded_file` (matches the newer logo handlers).

**XSS**
- `views/templates/hook/order_info.tpl` — escape `refundableQuantity` in the Payments-API refund button (the sibling Orders-API branch already escaped it).

**File security**
- Every shipped directory must contain an `index.php` to prevent directory browsing. The only directories missing one are the build-generated `vendor/**` and `views/js/admin/library/dist/**`, which aren't in git. Added an autoindex step to each packaging path (`Makefile:prepare-zip`, `create_zip.yml`, `release.yml`) that copies the module's `index.php` into any directory lacking one, right before the ZIP is assembled — so the shipped artifact has no browsable directories. No CI behaviour change.

## How to test

Deployed on a PrestaShop 9.0.1 shop with a Mollie test key:
1. **Admin** — API Configuration connects; Payment Methods lists methods fetched from the Mollie API; enabling iDEAL/Card saves without error (exercises the changed repositories).
2. **Checkout** — add a product, guest checkout, reach the payment step: enabled Mollie methods render (exercises `getMethodsForCheckout`). Methods with country/currency restrictions (e.g. iDEAL for a non-EUR order) are correctly hidden.
3. **Subscription ownership** — the guard rejects any `recurring_order_id` not owned by the logged-in customer before cancel / payment-method change.
4. **File security** — build the ZIP (`make prepare-zip` or the Create module ZIP workflow) and confirm `vendor/` and `views/js/admin/library/dist/` subdirectories contain an `index.php`.

All 11 code files pass `php -l` on PHP 8.4 and load cleanly under the PrestaShop kernel.

## Follow-ups (separate tickets)

- **PIPRES-807** — Apple Pay Direct AJAX unauthenticated cart IDOR (needs a session-binding refactor + Apple Pay device testing).
- **PIPRES-808** — webhook `security_token` / `secure_key` comparison and admin-AJAX CSRF hardening (needs Mollie webhook regression testing).

## Needs human verification

- The subscription ownership guard was verified by code review and logical equivalence to the existing `initContent()` ownership check; it was not exercised against real subscription data (the test shop had no subscriptions).
- Webhook flows were not run end-to-end (no live Mollie callbacks in the test environment); the two webhook items are deferred to PIPRES-808.

Confidence: 90%


TLabutis:
Merged develop into the branch to resolve conflicts. Two hunks were superseded by PRs that landed first and were resolved by keeping develop's side: the recurringOrderDetail ownership guard arrived via #1444 (its version also validates the id and customer object), and the order_info.tpl escaping targeted the old panel markup that the #1457 redesign replaced (the new template escapes its outputs). Your SQL casts, upload validation and index.php packaging changes are untouched. Details in the merge commit message.

TLabutis (approved):

PR #1492, opened 2026-08-25
Make the upgrade-check job boot its shop and actually test the upgrade

TLabutis:
The `Module-upgrading-check` job is supposed to install an old release, upgrade it to the code in the PR, and prove the upgrade worked. It was doing none of the three, and every defect was hidden behind the one before it.

### What was broken

**1. The shop never booted.** `docker-compose.1785.yml` hands `DB_PASSWD` to the PrestaShop container and `PS_INSTALL_AUTO` uses it against MySQL, whose root password is `prestashop`. The workflow never set it, so compose substituted an empty string, the auto-install could not authenticate and Apache never started. Every `prestashop:module ...` step then ran against nothing, returned in under 250 ms and exited 0. Locally the Makefile does `include .env`, which is why this only ever failed in CI.

**2. The tag checkout was partial.** `git checkout <tag> .` restores the tag's files but never deletes files added after it, so `config/services.yml` survived, kept declaring `SubscriptionFAQController`, and was compiled against an autoloader composer had just rebuilt from the tag's `composer.json`. Now a real `git checkout --detach --force`.

**3. The install was a no-op.** `e2eh1785` leaves the module installed, and `module install` on an installed module returns 0 without doing anything, so the shop never went back to the old release and the upgrade had nothing to upgrade. The target now uninstalls first, while the current code is still checked out and matches the schema it created.

**4. It upgraded the Addons release, not the PR.** The big one. PS 1.7.8 `ModuleManager::upgrade()` calls `setModuleOnDiskFromAddons()` before running any upgrade file, which downloads the published zip over `/var/www/html/modules/mollie` - a bind mount of the checkout. Locally this rewrote 1473 working-tree files in one second and left `changelog.md` holding the released blob. The console command has no `--source`, so it is replaced by `.docker/upgrade-module.php`, which boots the kernel (upgrade files reach for Symfony services, and `SymfonyContainer::getInstance()` needs a booted `$kernel`), reads `database_version` from `ps_module`, runs the local `upgrade/*.php` through `initUpgradeModule()` + `runUpgradeModule()`, and exits non-zero with the failing file and line instead of a bare fatal.

**5. The return leg had no composer.lock.** It is gitignored on the branch but tracked in the old tags, so the round trip deleted it and `composer install` silently became a full resolve that died on the anonymous GitHub rate limit. The head's lock is now kept aside, and the job sets `COMPOSER_AUTH` since it resolves twice.

**6. The old tag can no longer be resolved.** Its `require-dev` pins `roave/security-advisories dev-latest`, always today's advisory list, which now conflicts with every `guzzlehttp/psr7` the tag allows. `--no-dev` does not help, composer still solves require-dev. The tag is resolved with that package removed.

### What makes it a test now

The target asserts the shop's own `ps_module.version` at both ends, so a vacuous run fails instead of going green. It also returns to the commit under test rather than `develop`, and the base is `v6.0.1`, the oldest release in Mollie's compatibility table (PS 1.7.6 - 9.0.0 pairs with Mollie 6; there is no v6.0.0 tag, only betas). That pin needs bumping when Mollie 6 leaves support.

### Product bug this uncovered

`upgrade/Upgrade-5.3.0.php` called `Installer::createChargedbackState()`, removed in `74ee7c787` on 2023-07-18, so every merchant upgrading from below 5.3.0 hit a fatal. It now calls `OrderStateInstaller::install()`, the same thing `Upgrade-4.2.0.php` does, idempotent through `validateIfStatusExists()`. Changelog entry added under 6.4.6.

The same orphaned calls exist in `Upgrade-4.0.7.php` and `Upgrade-4.1.0.php`, left alone: Mollie 4 only ever supported PrestaShop 1.6.1, outside the module's `ps_versions_compliancy`.

### Verification

Run locally from `docker compose down -v`, on the v6.0.1 base:

```
E2E_EXIT=0
OK: shop reports module version 6.0.1
Upgraded mollie from 6.0.1 to 6.4.5, 13 upgrade file(s) applied.
OK: shop reports module version 6.4.5
UPGRADE_EXIT=0
```

Negative test on a v5.2.0 base, which does run `Upgrade-5.3.0.php`: with the fix it applies 17 upgrade files and lands on 6.4.5; with the old `createChargedbackState()` call put back the runner prints the file and line, exits 1, and `ps_module.version` stays at 5.2.0.

Note the workflow triggers on `pull_request: types: [opened, reopened]` only, so pushing to a branch does not re-run it.

### DIFF
--- a/.docker/upgrade-module.php
+++ b/.docker/upgrade-module.php
@@ -1,0 +1,94 @@
+<?php
+/**
+ * Runs a module's own upgrade/Upgrade-*.php files against the version the shop has
+ * registered, the way PrestaShop does once a merchant has replaced the module files.
+ *
+ * `bin/console prestashop:module upgrade` cannot be used for this. ModuleManager::upgrade()
+ * calls setModuleOnDiskFromAddons() whenever the marketplace advertises a newer version,
+ * which unpacks the released module over modules/<name> - here a bind mount of the
+ * checkout - so the code under test is thrown away before a single upgrade file runs.
+ *
+ * Usage: php .docker/upgrade-module.php <module>
+ */
+
+if (php_sapi_name() !== 'cli') {
+    exit(1);
+}
+
+$moduleName = isset($argv[1]) ? $argv[1] : '';
+
+if ($moduleName === '') {
+    fwrite(STDERR, "usage: php .docker/upgrade-module.php <module>\n");
+    exit(1);
+}
+
+require_once __DIR__ . '/../../../config/config.inc.php';
+
+$context = Context::getContext();
+
+if (!$context->employee) {
+    $context->employee = new Employee(1);
+}
+
+$query = new DbQuery();
+$query->select('version')
+    ->from('module')
+    ->where('name = "' . pSQL($moduleName) . '"');
+
+$registeredVersion = Db::getInstance()->getValue($query);
+
+if (!$registeredVersion) {
+    fwrite(STDERR, sprintf("The shop has no %s module installed.\n", $moduleName));
+    exit(1);
+}
+
+$module = Module::getInstanceByName($moduleName);
+
+if (!$module instanceof Module) {
+    fwrite(STDERR, sprintf("Could not instantiate the %s module.\n", $moduleName));
+    exit(1);
+}
+
+$module->installed = true;
+$module->database_version = $registeredVersion;
+
+if (!Module::initUpgradeModule($module)) {
+    fwrite(STDERR, sprintf(
+        "Nothing to upgrade: the shop reports %s %s and the files ship %s.\n",
+        $moduleName,
+        $registeredVersion,
+        $module->version
+    ));
+    exit(1);
+}
+
+$upgrade = $module->runUpgradeModule();
+
+foreach ($module->getErrors() as $error) {
+    fwrite(STDERR, strip_tags($error) . "\n");
+}
+
+if (!$upgrade['success'] || $upgrade['number_upgrade_left'] > 0) {
+    fwrite(STDERR, sprintf(
+        "Upgrading %s from %s failed on version %s, %d file(s) left.\n",
+        $moduleName,
+        $registeredVersion,
+        $upgrade['version_fail'],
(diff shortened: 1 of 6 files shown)


TLabutis:
### This PR's own run confirms the fix, and uncovers the next defect

`Module-upgrading-check` on this branch now gets **past** the point where it used to die:

```
Shop on port 8002 is up (HTTP 302); MySQL on 9002 accepts connections.
```

No `DB_PASSWD ... Defaulting to a blank string` warning, and the whole `make e2eh1785` target completes: seeding, module install/uninstall/install/enable, chmod. That is the failure in #1491 gone.

The job is still red, but on a **different and later** step, `make upgrading-module-test-1785`:

```
In RegisterControllerArgumentLocatorsPass.php line 68:
  Class "Mollie\Subscription\Controller\Symfony\SubscriptionFAQController"
  used for service "Mollie\Subscription\Controller\Symfony\SubscriptionFAQController"
  cannot be found.
```

This is pre-existing and unrelated to the change here. Cause:

```make
upgrading-module-test-$(VERSION):
	git checkout v5.2.0 .
	composer install
	docker exec ... prestashop:module install mollie
```

`git checkout <tag> .` restores the tag's tracked files but never deletes files added *after* the tag. v5.2.0 has `config/service.yml` (singular) and no `config/services.yml` (plural), so `develop`'s `config/services.yml` survives the checkout and keeps declaring the subscription controller - while `composer install` has just rebuilt the autoloader from v5.2.0's `composer.json`, which has neither the `Mollie\Subscription\` PSR-4 entry nor `subscription/` in the classmap. Symfony compiles the container, cannot resolve the class, and aborts.

So the "v5.2.0" leg of the upgrade test has never actually been a v5.2.0 tree. It went green only because the shop was never installed, so `bin/console` did nothing and exited 0 - the same fake green documented in the description.

Fixing that properly means reworking the target (a real checkout of the tag rather than a partial one, plus the `git checkout develop --force` issue noted in the description). Out of scope for this PR, which is about getting the shop to boot at all.

GytisZum (approved):

PR #1496, opened 2026-08-28
PIPRES-806 Copy payment method configuration from Test to Live

TLabutis:
PIPRES-806

Merchants configure every payment method in Test, switch to Live, and find an empty list. They then rebuild it by hand.

This adds a checkbox in the Test to Live switch dialog: **Copy my Test payment method settings to Live**. Off by default.

## What it does

- Copies per method: enabled flag, API choice, description, logo, min/max amount, manual capture, the full payment fee block, countries, excluded countries, excluded customer groups. Live order mirrors Test.
- Copies `MOLLIE_IFRAME` and `MOLLIE_SINGLE_CLICK_PAYMENT`, the only two configs with a separate value per environment.
- Skips methods not activated on the live account, and logs which ones (BO > Mollie > Logs).
- Never touches the API keys.
- One transaction, rolls back on any error. Re-runs update, never duplicate.

## Decisions worth a reviewer's attention

- **It replaces, it does not merge.** An existing Live method gets overwritten from Test, restrictions included. Safe because the checkbox is opt-in and off by default. Needs a product sign-off.
- **No standalone button** (ticket asked for one). Product chose the switch dialog instead.
- **Skipped methods are logged, not shown on screen.** Half of AC-2 met in the log. Putting it on screen is UI-only work.
- **The copy intent is not stored server side.** With no live key yet it waits in component state for the next connect. A stored flag could fire weeks later and overwrite Live settings built by hand in between.
- **Also fixes dropped margins on this page.** Tailwind's `@layer utilities` loses to the unlayered BO stylesheet, so every `mt-*`/`mb-*` on a `p`, `button` or `label` computed to `0px`. Fixed inline here; the global fix belongs in its own ticket.

## Verification

| | |
|---|---|
| Tests | `EnvironmentSettingsCopyServiceTest` 7 tests / 37 assertions. Unit suite 367/866. 2 pre-existing Integration errors, same on `develop` |
| CI | 12/12 green at `214926e` |
| Live account | Real `live_` key on PS 9.1.3: 24 copied, 6 skipped, field by field diff `OK=24 DIFF=0`. The 6 skipped are `pending-review` on the account, so the skip is correct |
| Browser | PS 1.7.6.1 and PS 9.1.3, 0 console errors, no 4xx/5xx. Checkbox default off, keyboard focusable, aligned at 375/768/1440. Rollback and zero-Test-rows paths both re-tested |
| API keys | Re-read after the live run, unchanged |
| Security | New AJAX action reads no request input; shop id comes from server context; no new SQL; auth and CSRF inherited from `ModuleAdminController`. No residual risk found |
| Changelog | Entry under `## Changes in release 6.4.6` |

Screenshot of the checkbox is attached by hand. QA artifacts are not committed to this repo, so there is no raw URL to pin.

## Needs human verification

1. Product sign-off on the replace semantics (point 1 above).
2. Product call on logging the skipped list instead of showing it.
3. Product call on dropping the queued copy when the page is reloaded before a key is connected.
4. PS 8 not opened in a browser for this page. CI's PS8 jobs pass.
5. No axe-core run, no contrast measurement. Keyboard checked by hand.
6. No e2e test. The repo has no back office API-configuration spec to extend.

Verdict: **ship with follow-ups**

Confidence: 90%

The one previously unproven path, the live Mollie call that drives the skip decision, has now run end to end with a real key. The remaining 10% is PS 8 browser coverage and the two product calls above.


TLabutis:
copy-checkbox-dialog

sarasko (approved):

PR #1501, opened 2026-08-31
PIPRES-820 Fix employee permission checkbox for the Mollie menu tab

TLabutis:
Fixes [PIPRES-820](https://mollie.atlassian.net/browse/PIPRES-820).

## The problem

Ticking View for the Mollie menu tab never sticks, so non-SuperAdmin profiles cannot get access to the Mollie menu. The grant is saved, but PrestaShop reads permissions back with a regex that cannot hold an underscore in the class name. The root tab is called `AdminMollieModule_MTR`, so `ROLE_MOD_TAB_ADMINMOLLIEMODULE_MTR_READ` parses as class `ADMINMOLLIEMODULE` plus action `MTR` and the grant lands on the hidden Settings tab instead.

This is a known core bug, identical on PS 1.7.8 / 8.x / 9.x: [PrestaShop/PrestaShop#28308](https://github.com/PrestaShop/PrestaShop/issues/28308), open since 2022 with no merged fix. The module has to absorb it.

## The fix

Two parts:

1. Rename the root tab to `AdminMollieModuleMTR` so its slugs have no underscore. `Upgrade-6.4.6.php` renames the tab and its four role slugs in place, keeping `id_authorization_role` and with it every grant merchants already made.
2. Pre-create the complete 16-tab `getTabs()` tree in `installSpecificTabs()`. The `_MTR` suffix is core's own: after install, `ModuleTabRegister::duplicateParentIfAlone()` clones any childless parent into a `<name>_MTR` dummy, which is what re-broke the tab tree on reinstall and turned the e2e jobs red on the first version of this PR. Core skips any class name already in `ps_tab`, so with the full tree pre-created it registers nothing and the dummy never comes back.

`Upgrade-6.4.0.php` deliberately keeps the old name: its delete list depends on it, and 6.4.6 renames right after it runs.

| File | What |
| --- | --- |
| `src/Install/Installer.php` | Root tab renamed; full tab tree pre-created; `deleteLegacyRootTab()` removes the leftover root tab that uninstall never deleted |
| `upgrade/Upgrade-6.4.6.php` | New. In-place rename of the tab and its role slugs |
| `views/css/admin/menu.css` | Selector follows the new class name (keeps the Mollie menu logo) |
| `mollie.php` | Version 6.4.6 so the upgrade script runs |
| `changelog.md` | 6.4.6 entry |

Ticket asks: the View checkbox must stay ticked and must grant menu access. Both confirmed by browser clicks, see Manual verification.

## Tests

| Check | Result |
| --- | --- |
| `php -l` on all changed PHP files | clean |
| php-cs-fixer dry run (`php:8.0-cli` container) | clean |
| grep for stale `AdminMollieModule_MTR` references | only the two deliberate legacy constants |

No unit test added: the defect lives in a core regex plus tab/role DB rows, so it is only observable against a real shop.

## Manual verification

- **PS 9.1.3, fresh install, browser**: View ticked for profile 4 stays ticked after reload, the role lands on `ADMINMOLLIEMODULEMTR_READ`, unticking removes it.
- **PS 9.1.3, the CI install-uninstall-install cycle**: both installs produce identical clean tab trees, no `_MTR` tab, zero underscore slugs. Browser check mirrors the two previously failing specs: all six `#subtab-*Parent` ids render and clicking Advanced Settings lands on the React page.
- **PS 8.2.3, real 6.4.5 to 6.4.6 upgrade** on the shop that reproduced the bug: tab and roles renamed in place, all four existing grants survive and read back, a profile 4 Salesman finally sees the Mollie menu.
- **PS 9.1.3, chained 6.3.4 to 6.4.6 upgrade**: one root tab, zero legacy slugs, grant/revoke round trip works.

Screenshots are in `~/Desktop/PIPRES-820/` (before/after of the checkbox on 6.4.5 and on this branch, plus the PS 8.2.3 post-upgrade grid); `gh` cannot attach images, I will drag them in from the web UI if needed.

## Needs human verification

- PS 1.7.8 upgrade path was not run locally (core code is byte-identical; the `local (PS1785)` e2e job covers the install cycle).
- `6.4.6` as the next version is a release manager call; changelog entry and upgrade file name follow it.
- Separate pre-existing defect, not fixed here: any 6.3.x to 6.4.x upgrade loses Mollie tab permissions because `Upgrade-6.4.0.php` deletes and recreates the tabs. Worth its own ticket.

Confidence: 92%

[PIPRES-820]: https://mollie.atlassian.net/browse/PIPRES-820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TLabutis:
CI caught a real regression from this branch, and it invalidates the approach. Holding this as draft.

`local (PS8)` and `local (PS1785)` fail on two specs, and they pass on #1490, #1494 and #1496, so this is not a flake:

- `install-uninstall.spec.ts:15 module tabs are all registered` - `#subtab-AdminMolliePaymentMethodsParent` resolved to 0 elements
- `advanced-settings.spec.ts:27` - timed out clicking `#subtab-AdminMollieAdvancedSettingsParent a`

(`Module-upgrading-check (PS1785)` also fails, but that one is the known-broken job from #1492. Everything else is green.)

## Reproduced

PS 9.1.3, running the `installing-uninstalling-enabling-module` cycle by hand. After the second install:

```
195   33  AdminMollieModuleMTR                 parent IMPROVE
204  195  AdminMollieModule_MTR      <-- reappeared, nothing in this branch installs it
196  204  AdminMollieModule                    parent AdminMollieModule_MTR
207  204  AdminMolliePaymentMethodsParent      parent AdminMollieModule_MTR
208  204  AdminMollieAdvancedSettingsParent    parent AdminMollieModule_MTR
197  195  AdminMollieAuthenticationParent      parent AdminMollieModuleMTR
```

Every tab declared in `mollie.php::getTabs()` landed under a resurrected `AdminMollieModule_MTR`, so the sidebar sub-tabs never render.

## Why: `_MTR` is generated by PrestaShop core

`src/Adapter/Module/Tab/ModuleTabRegister.php`:

```php
public const SUFFIX = '_MTR';
```

Identical in PS 1.7.8, 8.2.3 and 9.1.3.

- `duplicateParentIfAlone()` clones a parent tab and appends `_MTR` the first time a child is registered under a childless parent.
- `findParentId()` resolves a declared `parent_class_name` by looking up `$parentClassName . '_MTR'` **first**, plain name second. Core's parenting contract depends on that underscore.
- `ModuleTabUnregister::removeDuplicatedParent()` identifies the dummy parent by exactly `$parent->getClassName() === $child->getClassName() . SUFFIX` and collapses it on uninstall.

So `AdminMollieModule_MTR` was never a name this module chose. Core derived it from `AdminMollieModule`, and it got hardcoded into `Installer` and `Upgrade-6.4.0.php` afterwards. Renaming it means core stops recognising the tab as its own dummy parent, regenerates `AdminMollieModule_MTR`, and re-parents everything under that.

## Where that leaves PIPRES-820

The defect itself still stands. `Profile::generateAccessesArrayFromPermissions()` matches with:

```
/ROLE_MOD_[A-Z]+_(?P<classname>[A-Z][A-Z0-9]*)_[A-Z]+/
```

which cannot parse a class name containing an underscore. But since the underscore comes from core's own `SUFFIX`, this is a PrestaShop core bug that hits every module whose root menu tab gets the `_MTR` treatment, not a Mollie naming mistake. The rename cannot be the fix on its own.

Not converting this to ready for review. Needs a decision on approach first.

TLabutis:
Fixed in b2d215387. The rename is kept and installSpecificTabs() now pre-creates the complete 16-tab getTabs() tree, so core's ModuleTabRegister skips every class name via checkIsValid() and duplicateParentIfAlone() never regenerates the _MTR dummy. Verified locally on PS 9.1.3 with the exact CI install-uninstall-install cycle (identical clean trees both times) plus a browser check mirroring both failing specs. Note: the 9.1.3 env mentioned above no longer holds the duplicated tree, it was consumed by this retest. Upstream context: this is PrestaShop core bug PrestaShop/PrestaShop#28308, open since 2022 with no merged fix.

sarasko (approved):

Issue #1503, opened 2026-09-02
PHP Deprecated errors


owebia:
Hi,

With version 6.4.5 of the module, I get these deprecated errors on the dependencies:

```
PHP Deprecated:  Invertus\Knapsack\filter(): Implicitly marking parameter $function as nullable is deprecated, the explicit nullable type must be used instead in modules/mollie/vendor/invertus/knapsack/src/collection_functions.php on line 248
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in modules/mollie/vendor/segmentio/analytics-php/lib/Segment.php on line 112
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in modules/mollie/vendor/segmentio/analytics-php/lib/Segment/Consumer/ForkCurl.php on line 47
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in modules/mollie/vendor/segmentio/analytics-php/lib/Segment/Consumer/ForkCurl.php on line 82
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in modules/mollie/vendor/segmentio/analytics-php/lib/Segment/Consumer/LibCurl.php on line 72
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in modules/mollie/vendor/segmentio/analytics-php/lib/Segment/Consumer/Socket.php on line 182
```

Could you please update the dependencies?

fjbender:
Could be related to #1410 ? Which PHP version are you using?

owebia:
PHP 8.5

fjbender:
Clear, thanks - in this case I'd like to treat this as a duplicate of #1410 which we're internally tracking as PIPRES-814 and have scheduled a fix for version 6.4.6 currently planned for middle of this month.

owebia:
Thanks! Will the fix also work for PHP 8.5 (more deprecated syntax than PHP 8.3)?

fjbender:
Yes, the syntax target is 8.5

owebia:
Perfect, thanks @fjbender !

PR #1509, opened 2026-09-04
PIPRES-814: keep null and untyped jsonSerialize off the payment path

sarasko:
Ticket: [PIPRES-814](https://mollie.atlassian.net/browse/PIPRES-814)

## Summary
- Stop handing `null` to `number_format()` when building Mollie order lines, and declare the return type intent on the four `jsonSerialize()` implementations that declared none.
- Why: both are deprecations on PHP 8.1+ today and hard errors on PHP 9, so the payment step of checkout would stop working the day a merchant's host upgrades. Nothing is visible to the shopper today beyond log growth.
- Redone on `develop` from community PR #1411, which sits on the retired `release-6.4.4` branch. The two `package-lock.json` changes it carried are dropped.
- Scope is the core fix only, per the decision on the ticket. The PHP 9 readiness work (the implicitly-nullable parameter defaults) is deliberately not here.

## Proof
Deprecations raised by module code on the payment path, same module tree, before and after the fix:

| PHP | before | after | `formatNumber(null, 0)` |
|---|---:|---:|---|
| 7.2.34 | 0 | 0 | `"0"` |
| 7.4.33 | 0 | 0 | `"0"` |
| 8.1.33 | 5 | 0 | `"0"` |
| 8.3.28 | 5 | 0 | `"0"` |
| 8.5.10 | 5 | 0 | `"0"` |

Unit suite on the CI runtime (PHP 7.4 + the pinned phpunit 7.5):

```
......                                                              6 / 6 (100%)
OK (6 tests, 7 assertions)
```

`phpstan analyse --configuration=tests/phpstan/phpstan_base.neon` against PrestaShop 1.7.6.8: `[OK] No errors`. php-cs-fixer: no changes to make.

Browser QA does not apply here (`delivery-playbook classify` reports `Browser QA: N/A`, no UI files). The payment path was exercised server-side instead, against a real cart on a running PrestaShop 9.2 shop: see Details.

The ticket carries no `AC-<n>` ids. These criteria are taken from the ticket description and the scoping decision in its comments.

- **AC-1** A checkout on PHP 8.1+ raises no `number_format(): Passing null` deprecation from module code
  - `TextFormatUtilityTest::testFormatNumberEmitsNoDeprecationForNull`
  - real-cart run on a live shop, in Details
- **AC-2** No `jsonSerialize()` return-type deprecation from module code
  - the before/after table above, reproduced on PHP 8.1, 8.3 and 8.5
- **AC-3** `formatNumber(null, 0)` still returns `"0"`, so no output changes
  - `TextFormatUtilityTest::testFormatNumber`, case `null price formats as zero`
- **AC-4** Code stays compatible with the PHP 7.2.5 composer platform pin
  - the PHP 7.2.34 and 7.4.33 rows above: all four classes load and behave identically
- **AC-5** A unit test pins the null case and fails on the parent commit
  - verified, output in Details
- **AC-6** No lock-file or dependency changes in the diff
  - `git diff origin/develop --stat`: 6 files, all source or test

## Verify before merge
- Community PR #1411 is still open. Close it in favour of this branch on merge, along with issue #1410.
- Issue #1503 is referenced on the ticket as the same defect. It is not. Every line it reports is in `vendor/` (`invertus/knapsack`, `segmentio/analytics-php`) and needs a dependency bump, not this fix. It needs its own ticket and is not addressed here.

<details>
<summary>Details</summary>

### Root cause

`CartLinesService::createProductLines()` builds a product hash per cart line:

```php
$idCustomization = TextFormatUtility::formatNumber($cartItem['id_customization'], 0);
```

`Cart::getProducts()` returns `id_customization => null` for any product without a customisation, so an ordinary checkout reaches `number_format(null, ...)`. Lines 190 and 191 pass `id_product` and `id_product_attribute`, which are always set, so this one line is the trigger.

The cast lands in `formatNumber()` rather than at the call site. That is what PR #1411 does, it covers all sixteen call sites at once, and it changes no output: `number_format(null, ...)` already returned `"0"`.

The four `jsonSerialize()` implementations without a return type were `DTO/Object/Payment`, `DTO/Object/Company`, `DTO/ApplePay/Carrier/Carrier` and `Errors/Error`. The other eight in the module already declare `: array` or `: string`, which are covariant with `mixed`, so they are untouched. A sweep of the module found no further occurrence of either pattern.

`#[\ReturnTypeWillChange]` was chosen over declaring `: mixed` because `: mixed` needs PHP 8.0 and the composer `platform` pin is 7.2.5. PHP 7 parses the attribute as a comment, which the 7.2.34 row above confirms.

### The test failing on the parent commit

Same test, same PHP 8.5, with only `src/Utility/TextFormatUtility.php` reverted:

```
1) Mollie\Tests\Unit\Utility\TextFormatUtilityTest::testFormatNumberEmitsNoDeprecationForNull
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
-Array &0 []
+Array &0 [
+    0 => 'number_format(): Passing null to parameter #1 ($num) of type float is deprecated',
+]
```

### Real cart on a running shop (PrestaShop 9.2, PHP 8.3)

`CartLinesService::getCartLines()` called with a real cart's `getProducts()`, `getSummaryDetails()` and shipping cost:

```
before:  item 0 id_product=1 id_product_attribute=1 id_customization=NULL
         order lines built: 2
         DEPRECATIONS: 1
           - number_format(): Passing null to parameter #1 ($num) of type float is deprecated   @ src/Utility/TextFormatUtility.php:24

after:   item 0 id_product=1 id_product_attribute=1 id_customization=NULL
         order lines built: 2
         DEPRECATIONS: 0
```

Same cart, same two order lines built, so the fix removes the deprecation without changing what is sent to Mollie.

### Why no committed test covers the jsonSerialize half

That deprecation is raised at class declaration time and only on PHP 8.1+. CI runs the unit suite on PHP 7.4 (`make start-ps-for-tests` boots PrestaShop 1.7.8 / PHP 7.4), where it cannot be observed at all, so a committed test would pass whether or not the fix is present. It is proven by reproduction on PHP 8.1, 8.3 and 8.5 instead.

The durable guard is a CI job on a modern PHP that fails on new deprecations. That is already scoped into the deferred PHP 9 readiness item rather than added here.

</details>

Confidence: 4/5 - every criterion is backed by a named test or a before/after reproduction across five PHP versions, and phpstan and php-cs-fixer are clean; one criterion rests on reproduction rather than a committed test, because the PHP 7.4 CI runtime cannot observe that class of deprecation.

[PIPRES-814]: https://mollie.atlassian.net/browse/PIPRES-814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DTO/Object/Payment.php
+++ b/src/DTO/Object/Payment.php
@@ -108,6 +108,7 @@ class Payment implements \JsonSerializable
         $this->company = $company;
     }
 
+    #[\ReturnTypeWillChange]
     public function jsonSerialize()
     {
         $result = [];
--- a/src/DTO/Object/Payment.php
+++ b/src/DTO/Object/Payment.php
@@ -108,8 +108,7 @@ class Payment implements \JsonSerializable
         $this->company = $company;
     }
 
-    #[\ReturnTypeWillChange]
-    public function jsonSerialize()
+    public function jsonSerialize(): array
     {
         $result = [];
         $result['cardToken'] = $this->getCardToken();
--- a/src/DTO/ApplePay/Carrier/Carrier.php
+++ b/src/DTO/ApplePay/Carrier/Carrier.php
@@ -68,6 +68,7 @@ class Carrier implements JsonSerializable
         return $this->amount;
     }
 
+    #[\ReturnTypeWillChange]
     public function jsonSerialize()
     {
         return [
--- a/src/DTO/ApplePay/Carrier/Carrier.php
+++ b/src/DTO/ApplePay/Carrier/Carrier.php
@@ -68,8 +68,7 @@ class Carrier implements JsonSerializable
         return $this->amount;
     }
 
-    #[\ReturnTypeWillChange]
-    public function jsonSerialize()
+    public function jsonSerialize(): array
     {
         return [
             'label' => $this->getName(),
--- a/src/DTO/Object/Company.php
+++ b/src/DTO/Object/Company.php
@@ -49,6 +49,7 @@ class Company implements \JsonSerializable
         $this->registrationNumber = $registrationNumber;
     }
 
+    #[\ReturnTypeWillChange]
     public function jsonSerialize()
     {
         $json = [];
--- a/src/DTO/Object/Company.php
+++ b/src/DTO/Object/Company.php
@@ -49,8 +49,7 @@ class Company implements \JsonSerializable
         $this->registrationNumber = $registrationNumber;
     }
 
-    #[\ReturnTypeWillChange]
-    public function jsonSerialize()
+    public function jsonSerialize(): array
     {
         $json = [];
         $json['vatNumber'] = $this->getVatNumber();
--- a/src/Errors/Error.php
+++ b/src/Errors/Error.php
@@ -40,6 +40,7 @@ class Error implements \JsonSerializable
         return $this->message;
     }
 
+    #[\ReturnTypeWillChange]
     public function jsonSerialize()
     {
         $json = [];
--- a/src/Errors/Error.php
+++ b/src/Errors/Error.php
@@ -40,8 +40,7 @@ class Error implements \JsonSerializable
         return $this->message;
     }
 
-    #[\ReturnTypeWillChange]
-    public function jsonSerialize()
+    public function jsonSerialize(): array
     {
         $json = [];
         $json['code'] = $this->code;
--- a/src/Utility/TextFormatUtility.php
+++ b/src/Utility/TextFormatUtility.php
@@ -21,7 +21,7 @@ class TextFormatUtility
 {
     public static function formatNumber($unitPrice, $apiRoundingPrecision, $docPoint = '.', $thousandSep = '')
     {
-        return number_format($unitPrice, $apiRoundingPrecision, $docPoint, $thousandSep);
+        return number_format((float) $unitPrice, $apiRoundingPrecision, $docPoint, $thousandSep);
     }
 
     /**
--- a/tests/Unit/Utility/TextFormatUtilityTest.php
+++ b/tests/Unit/Utility/TextFormatUtilityTest.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * Mollie       https://www.mollie.nl
+ *
+ * @author      Mollie B.V. <info@mollie.nl>
+ * @copyright   Mollie B.V.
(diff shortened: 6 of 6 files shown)


src/DTO/Object/Payment.php
```diff
@@ -108,6 +108,7 @@ public function setCompany(Company $company): void
         $this->company = $company;
     }
 
+    #[\ReturnTypeWillChange]
```
GytisZum: This is introduced since PHP 8.1. Other PHP versions might break. Double check it

GytisZum (changes requested):

src/DTO/Object/Payment.php
sarasko: Good catch, and it pushed me to a better fix. I have dropped the attribute and declared the return type instead, in e6f4467.

All four methods return an array unconditionally, so the type can be stated outright:

```php
public function jsonSerialize(): array
```

`: array` is valid from PHP 7.0, and `array` is covariant with the `mixed` that `JsonSerializable` declares from 8.1, so the deprecation is silenced with no attribute and no version-conditional behaviour. It also matches the eight `jsonSerialize` implementations in the module that already declare `: array` or `: string`.

On your original point: the attribute would not have broken older versions, because `#[` opens a `#` comment on PHP 7.x and PHP only resolves an attribute's class when it is instantiated through reflection, which nothing here does. I checked 7.2.34, 7.4.33, 8.0.30, 8.1.34, 8.3.33 and 8.5.10 and all four classes loaded and serialised on each. But it did rely on the attribute staying on one line, and it was a temporary suppression rather than a real declaration, so the return type is the better answer.

Verified after the change on 7.2.34, 7.4.33, 8.0.30, 8.1.34, 8.3.33 and 8.5.10: all four classes load, `json_encode` returns the expected payload, and no deprecation is raised from module code. Unit tests pass on the CI runtime, php-cs-fixer is clean, and PHPStan reports no errors against 1.7.6.8.

Note that `: mixed` was not an option here. It needs PHP 8.0 and `composer.json` pins the platform to 7.2.5.

TLabutis (approved):

TLabutis (approved):

GytisZum (approved):

GytisZum (approved):

